PR #55, opened 2017-09-08
Initial Directory

chigby:
This pull request:

1. Adds a `directory` app to contain the a list of known (and graded) SecureDrop instances and a way for users to submit their own.

2. Adds models for:
 * A `DirectoryPage`, a wagtail page that will contain the directory contents.
 * A `SecureDropInstance` -- a plain-old Django model containing metadata about an instance of SecureDrop.
 * A `Result` -- a Django model containing the results of a scan of a `SecureDropInstance`. This is largely the same as the `Result` object from the securedrop.party prototype.

3. Ports over an adaptation of the scanning code from the aforementioned prototype.

4. Adds a form to `DirectoryPage`s that lets the user submit information about their SecureDrop page. This is extremely rudimentary, but the "full stack" functionality of submitting a form and having the `SecureDropInstance` populate and a scan initiated and its `Result` saved is implemented. 

5. Very little on the front end. The DirectoryPage doesn't have much in the way of a template and doesn't display any data. This can and should be implemented once more of the design decisions have been finalized.

Refs #4


conorsch (changes requested):
Looks like this code is from https://github.com/freedomofpress/securedrop-landing-page-checker/, which is good! However, we've lost all of the commit history due to the copy/paste integration into this repository.

You should use `git subtree` to pull in the changes via PR so that 1) @redshiftzero gets credit for the early work via git metadata; and 2) we preserve history going back in time, so that future maintainers will have access to her helpful commit messages.

PR #423, opened 2018-03-23
Upgrade Django from 1.11.5 to 1.11.11

harrislapiroff:
Does what it says on the tin


conorsch (approved):
LGTM. Let's keep an eye on the typing dependency, but otherwise going in, will get into staging shortly.

PR #431, opened 2018-04-06
Add CSP for securedrop.org

emkll:
Adds an initial CSP to securedrop.org website using  django-csp to requirements and middleware.

CSP is mostly a defense-in-depth solution, and shouldn't break functionality. We can slowly iterate on this policy:

- The big win here is that we disable `unsafe-inline` JavaScript.
- Hashes are used for white listing inline style, which shouldn't break site functionality but could break appearance on upgrades. We should ensure there are no CSP style errors on upgrades.
- `unsafe-eval` is required for client/common/js/common.js:645 and /client/tor/js/torEntry.js:89. We should consider fixing this, as it would be another big win.

We should also investigate using a `report-uri` (to which CSP violations are reported to) to both detect errors in the policy, and be aware of potential attack vectors (obviously while preserving user privacy and in line as the currently level of logging)

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -14,6 +14,7 @@ defusedxml==0.5.0
 django-allauth-2fa==0.4.4
 django-allauth==0.34.0
 django-anymail[mailgun]==1.4
+django-csp==3.4
 django-debug-toolbar==1.9.1
 django-modelcluster==3.1
 django-otp==0.4.1.1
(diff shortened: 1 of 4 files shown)


conorsch (approved):
Let's toss this at staging and bang on the hood. @harrislapiroff, will wait for your +1 before merge and deploy.

PR #438, opened 2018-04-13
Directory restructure

harrislapiroff:
There's more to do for this restructure, but this comprises a fairly self contained PR. It does the following:

* Move `Result`, `SecuredropPage`, and `SecuredropOwner` from the `landing_page_checker` to the `directory` app
* Rename `SecuredropPage` to `DirectoryEntry`
* Move `DirectorySettings` from `common` to `directory` app
* Rename `landing_page_domain` fields to more accurate `landing_page_url` fields
* Move a bunch of affiliated code

Some notes:

* I did a bunch of work to make sure migrations run properly from the current database state as well as from scratch, but that should be verified
* Despite the above, this PR does add a bunch of migrations _and_ substantially increases the complexity of our migration tree. I might like to go through a procedure of resetting migrations before the actual site launch so that there's only one migration in each app at the time of launch. I know what this process needs to look like and can sketch it out later.

Things to do after this is merged:

* Restructure `accounts` and `landing_page_checker`. One should become a scanner module that _just_ includes scanner logic and is completely divorced from the Django ORM. The other should be the app that provides the directory management tools for securedrop instance admins.


conorsch (approved):
Super clear transition here. As you mention, there's still more reorg work to be done, particularly the old landing_page_checker dir, but your plan for next steps sounds great.

Issue #443, opened 2018-04-18
Investigate inconsistent test failure


harrislapiroff:
As detailed in https://github.com/freedomofpress/securedrop.org/pull/431#issuecomment-382420512

PR #445, opened 2018-04-18
Further scanner and directory restructure

harrislapiroff:
This PR will do the following things:

* Deconstruct the `landing_page_checker` app until there is nothing in it besides historical migrations
* Convert taxonomy models (`Language`, `Topic`, `Country`) from a concrete inheritance model to an abstract inheritance model, probably resolving #443
* Rename `Result` to `ScanResult`
* Establish a new `scanner` app to hold scanning logic
* Move directory management logic to `accounts` app. This may not be an obvious change, but I think it will make development going forward more straightforward since if we end up scrapping directory self-management, we'll also scrap user-facing auth. If we don't end up scrapping it, I'd expect user-facing auth and directory self-management to share a lot of views.
* Make taxonomy models editable from the Wagtail admin

This should be fairly easy to review commit by commit, but I'm also happy to separate certain items out into different PRs if that's preferred.

There's still more work that could be done here. In particular, I've previously mentioned decoupling the scanner logic from the django model logic—which will become necessary as we expand or reuse the scanner code—but I'm comfortable putting that off until after launch and would rather merge this now and focus on higher priority items.


conorsch (approved):
:sunglasses: :shipit:

PR #453, opened 2018-04-20
Ensure that log in link is hidden if directory tools are disabled

### LINKED ISSUE #430: Remove "Log In" link in footer when "Allow directory management" is disabled
On the home page footer, there is a Log In link, even when Allow directory management is disabled (default behavior in dev environment). This results in some confusion, as they have an account but can't do anything with it.
screenshot from 2018-04-06 14-29-20

I suggest that link be removed from the footer to reduce user confusion. Admins will still be able to login via /admin

harrislapiroff:
Also fix some other footer template issues in the process

To test:

* Toggle settings in the wagtail admin under Settings > Directory Settings > Allow Directory Management
* Observe that the footer "Log In"/"Manage your SecureDrop directory listing" link toggles on and off with the setting (this link is on every page)

Resolves #430


conorsch (approved):
Works as advertised! Confirmed via local testing.

PR #459, opened 2018-04-23
456 form aesthetic

### LINKED ISSUE #456: Form aesthetic
Make the form pages prettier!

harrislapiroff:
Improve form styles

* Remove special styles for django allauth. We'll implement something new for that if we ever move forward with self-management tools. Otherwise those styles are moot.
* Rewrite css styles
* Add a template tag to assist in styling different input types

New forms looks like this:

screen shot 2018-04-23 at 16 42 44

Resolves #456


conorsch (changes requested):
Requesting more beautified Submit button.

conorsch (approved):

PR #461, opened 2018-04-24
Fix some references from common.DirectorySettings to directory.DirectorySettings

harrislapiroff:


conorsch (approved):
Visual review LGTM. Inclined to toss at staging and confirm fix there.

PR #464, opened 2018-04-24
Directory entry logo improvements

### LINKED ISSUE #428: Enlarge logo on instance view page in directory
The logo appearance on the instance pages is very small right now, which is a problem esp. for logos with text in them, e.g.:

cpiexample

I suggest increasing the size by 30-50%, but @harrislapiroff feel free to choose whatever you think is a reasonable approach here. Here are a few different example sizes:

- square: https://staging-admin.securedrop.org/directory/associated-press/ (looks OK)
- banner text: https://staging-admin.securedrop.org/directory/buzzfeed/ (bit small)
- banner text (long): https://staging.securedrop.org/directory/barton-gellman-century-foundation/ (too small)
- banner text+graphic: https://staging.securedrop.org/directory/usa-today-network/ (too small)

harrislapiroff:
Directory entry logos can now expand to be up to 400px wide or 160px tall

I also added an option in the admin to specify that a logo _is_ the organization name, since we already have a number of logos that are designed this way. If we want to more fully embrace this, later on we could separate "square logo" and "name logo" into separate optional variants. With that option enabled a directory page now looks like this:

screen shot 2018-04-24 at 15 32 12

I also added a `srcset` attribute for high density display optimization

Resolves #428


conorsch (approved):
LGTM, will toss at staging shortly...

PR #469, opened 2018-04-25
Ignore release candidates

### LINKED ISSUE #468: Don't update latest release if the "release" is a release candidate
Today we're working on preparing the first release candidate for SecureDrop 0.7. However, my understanding is that GitHub treats all tags as Release objects and as such `rc` tags will update the latest release on the website. As such, we'd want to modify the latest release object to _only_ update when the tag does not contain `rc`. 

Note that we can wait to push tags and instead build test packages from commit hashes so this is not extremely urgent nor should it block launch.

harrislapiroff:
Resolves #468


conorsch (approved):
:+1: :shipit:

PR #479, opened 2018-04-26
Indexes and Search Optimization

### LINKED ISSUE #478: Throw indexes at some fields
One way we might be able to get some database performance improvements with minimal effort is to throw some indexes at fields that we query and order by regularly.

harrislapiroff:
Started this PR just adding indexes to ordering/filtering fields on models and ended up restructuring the search model a fair bit.

Performance gains on the indexes are minimal, but the performance gains on search brought queries from 3 seconds to .5 seconds.

Before | After
-------|------
screen shot 2018-04-26 at 12 56 05 pm | screen shot 2018-04-26 at 12 55 24 pm

One thought is that with these changes it would be possible to get rid of `Search.search_content` altogether and just allow the `search.utils.index_*` methods to write `SearchVector`s directly to the `search_vector` field. Advantages of that would be:

1. We'd go back to two sources of truth (the actual object, the `search_vector` field) instead of the three we have now (the actual object, the `search_vector` field, the `search_content` field)
2. The utility functions could define custom weights for various properties (for instance, maybe we want to weight the author name of a forum post more highly than the content)

This would be a larger restructure though, so if people think that would be a good approach maybe we can open a new ticket for it.

Resolves #478


conorsch (approved):
:shipit:

PR #519, opened 2018-06-06
 Detect Quantcast, comScore, Chartbeat, Krux scripts

### LINKED ISSUE #505: Extend landing page scanner analytics check beyond Google Analytics
This is a longer term improvement beyond the scope of the initial scanner integration (#488). Currently the scanner only checks for the presence of Google Analytics (see https://github.com/freedomofpress/securedrop.org/blob/master/scanner/scanner.py#L174). Let's audit the current set of news sites and add any other common analytics scripts we're missing. They would trigger the same severe warning (#496).

eloquence:
Fixes #505. These are the most common analytics scripts on news sites
in the SecureDrop directory.

# Testing

- [x] Verify that the newly added analytics scripts are the expected ones via the provided documentation. For comScore, I was not able to find official docs beyond the ones linked, but other sites also identify `beacon.js` as the relevant script, e.g., https://segment.com/docs/destinations/comscore/#analytics-js
- [x] Add a directory entry that has a landing page with none of the detected analytics scripts (e.g., `https://www.nytimes.com/newsgraphics/2016/news-tips/index.html`). Run the scanner (` docker exec sd_django /bin/bash -c "./manage.py scan"`) and verify that `No analytics` is set to `true` in the most recent results appended to the directory entry.
- [x] Add an entry with a landing page that has one of the detected analytics scripts, e.g., `https://www.nrk.no/` (not a landing page but it will do for the purpose of testing). Run the scanner and verify that `No analytics` is set to `false` in the most recent results appended to the directory entry.

I have performed these tests myself (as well as running the existing automated tests), but ideally the reviewer should verify this as well.


harrislapiroff (approved):
This looks good and seems to work properly. I'm approving it tentatively, but I wonder if we should update the scanner tests to include this criteria. @eloquence if you wanted to try to write a test, it might be useful to learn about [vcrpy](https://github.com/kevin1024/vcrpy). I'd also be happy to try my hand at a test when I'm further along with donation reconciliation.

chigby (approved):
I ran through the tests described, created a few directory entries and tested out the scanner against them and everything worked as I expected it to. I like the VCR-based tests and the code is simple enough to understand easily what is going on. So, everything in the PR itself is looking great.

At a higher level, I wondered if this was the best way to ensure that the landing page is not running any analytics/tracking software without actually executing the page's javascript and looking at the network traffic. I imagine this would require a more advanced and complex tool like Phantom.js. But I also remembered that the landing pages are run by news organizations we're working with, and the scanning tool is mostly to help us figure out that they have properly configured everything and not, say, forgotten to remove their comScore code from the template they use for the landing page. So I think the way we're doing it here suits that purpose just fine. Just thinking out loud, really.

conorsch (approved):
Reviewed visually, LGTM. Have not tested locally, but based on input from @harrislapiroff and @chigby , this looks ready to rock. Will queue up for a deploy later today with other outstanding changes pending merge.

PR #520, opened 2018-06-25
Perform nightly security checks in CI

emkll:
- Ensures the latest version of `bandit` and `safety` are installed each time CI runs
- De-couple `safety_check` workflow which will be run daily. It will also be run on each PR, in conjunction with the `build` workflow.


conorsch (approved):
LGTM. Will adjust the branch protection settings post-merge.

PR #523, opened 2018-07-16
Updates dev/CI dependencies Molecule 2.4 -> 2.13

conorsch:
We were seeing failing CI errors during cron checks due to the older
versions of Molecule failing during pull from pypi. Bumped to a more
recent version, should still be compatible with our configs as-is.

CI passing should be considered sufficient for merge.

Partial progress towards https://github.com/freedomofpress/fpf-www-projects/issues/18.


harrislapiroff (approved):
🕹✅

PR #527, opened 2018-08-07
Add scanresultcsv command

harrislapiroff:
#### To Test

1. Ensure you have a database with at least some directory entries and scan results

2. Run this management command through docker

   ```bash
   docker exec -it sd_django ./manage.py scanresultcsv
   ```

   and observe that a csv is output to stdout

3. Run the management command with the output file flag

   ```bash
   docker exec -it sd_django ./manage.py scanresultcsv -o output.csv
   ```

   and observe that an accurate csv is written to a file

### DIFF
--- a/directory/management/commands/scanresultcsv.py
+++ b/directory/management/commands/scanresultcsv.py
@@ -1,0 +1,27 @@
+import argparse
+
+from django.core.management.base import BaseCommand
+
+from directory.models.entry import DirectoryEntry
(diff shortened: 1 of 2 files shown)


conorsch (approved):
LGTM! Verified stdout and file outputs. Used dummy data from local dev env, so many URLs were not valid, but the fields are nonetheless populated accurately.

PR #528, opened 2018-08-17
Add csp report URI for securedrop.org

emkll:
Adds the CSP report uri for securedrop.org

Now that the freedom.press uri works, we can test it with securedrop.org in our next staging deployment. Report URI also allows us to filter by domain name.


conorsch (changes requested):
Two changes:

* more explicit string setting
* support env var

will edit and push up

conorsch (approved):

PR #535, opened 2018-09-06
Tor Browser 8 fixes

harrislapiroff:
This PR includes two commits:

####  Improve Tor detection script

Tor Browser 8 and Firefox Quantum have indistinguishable user
agent strings. `navigator.mimeType.length` is also no longer a
way to distinguish between Tor Browser and Firefox (both return 0).

Instead we now use a combination of:

- UA string
- GMT timezone
- Reported screen size and window size being identical

With the current versions of major web browsers, these factors
measure a strong likelihood of whether a browser is or isn't
Tor Browser

#### Fix Tor SVG fallback

* Use a CSS-hide/show based approach instead of the previous
  <object> tag approach, since Tor 8 now blocks <object> tags
  without displaying their fallback
* Catch all files that weren't using the central _svg.html template
  and update them to use it
* Remove an extraneous django template tag from an SVG file

An unfortunate downside of this approach is that the PNG files will
still load in browsers that support SVG, eliminating any
bandwidth-saving benefits of SVGs. It would be nice to find an
approach in the future that did not suffer this problem


conorsch (approved):
Local testing is promising, let's get it out.

PR #629, opened 2019-06-21
Ensure that frontend caches don't cache form pages

harrislapiroff:
CSRF tokens are distributed on form pages in the response on a per
user basis. A frontend cache like cloudflare can interfere with
this process by displaying the same token to multiple users due to
caching the response. The form then throws an error when the token
doesn't validate.

This change specifies that form pages are "private"--only to be
cached by the end-user's browser, not intermediaries. With this
change, cloudflare should not store form page responses at all,
resolving the CSRF issue.

See:

- https://support.cloudflare.com/hc/en-us/articles/115003206852-Enabling-Origin-Cache-Control-with-Cloudflare-Page-Rules
- https://docs.djangoproject.com/en/2.2/ref/csrf/

### DIFF
--- a/forms/models.py
+++ b/forms/models.py
@@ -1,4 +1,6 @@
 from django.db import models
+from django.utils.decorators import method_decorator
+from django.views.decorators.cache import cache_control
 from modelcluster.fields import ParentalKey
 from wagtail.admin.edit_handlers import (
     FieldPanel, FieldRowPanel,
@@ -37,6 +39,7 @@ class FormField(AbstractFormField):
     ]
 
 
+@method_decorator(cache_control(private=True), name='serve')
 class FormPage(MetadataPageMixin, AbstractEmailForm):
     intro = RichTextField(blank=True)
     warning = RichTextField(blank=True, help_text='A warning for sources not to submit documents via this form.')


harrislapiroff:
Once this is approved, we should make this change on other repos that also use form pages or csrf tokens.

harrislapiroff:
@conorsch and @ketudb are working out details here, but the plan is to put this on the sandbox server, ensure that cloudflare caching is turned on, and run some manual testing there before merging.

harrislapiroff:
I tested this on the sandbox by creating a test form at https://sandbox-upgrade.securedrop.org/test-form/. I was able to observe:

1. Cloudflare doesn't seem to munge the `private` declaration in particular. That made its way all the way through to the end response, which is perfectly fine.
2. Different csrf tokens were served in the HTML to different sessions (you can find this by searching for a hidden input with the name `csrfmiddlewaretoken`.
3. I was able to submit forms from two different sessions in rapid succession.

Will wait for @eloquence's 💯 before merging.

eloquence:
I can confirm your results, @harrislapiroff. New CSRF tokens are assigned for each view, form submissions succeed, and the `Cache-Control` directive is passed through unmolested. 

Given the survey in-progress, I'm still a bit paranoid about making form-related changes right now, but as long as we test immediately after the deploy and are prepared to revert, I'm OK with this going out in the next few days. :)

harrislapiroff:
Okay! I'm going to merge it. Given that it's just changing an HTTP header I would expect the odds of it interfering with form submission are slim, but will make sure @ketudb knows to give us a heads up when they deploy.

PR #640, opened 2019-08-06
Enhance Circle runs - DRYify, extract logs

msheiny:
# Description

Fixes https://github.com/freedomofpress/fpf-www-projects/issues/42 (for securedrop.org)

This commits performs the following:
* Fixes the `machine` field to match upstream reference spec
* Takes advantage of yaml anchor tags to DRYify some of the python deps
* Gathers docker logs at the end of a test for developer debug assistance

# Testing
* Does CI pass?
* Can you see docker logs in the artifacts tab under both dev / prod run environments?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,20 +1,30 @@
 ---
+reuse-blerbs:
+  - &python_prereqs
+    run:
+      name: Install testing pre-reqs
+      command: |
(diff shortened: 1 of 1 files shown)


msheiny:
`ci/circleci: build` is not going to pass - im updating the repo behind the scenes to swap the required checks here but ignore that during review

PR #654, opened 2019-12-13
Upgrade to latest version of npm and supported version of node

harrislapiroff:
* In the process I've also extended our NodeDockerfile to allow
  us to easily specify an NPM version as part of the
  docker-compose file
* Since we are still on a supported LTS track with Node 10.x, I
  decided to conservatively upgrade to the latest version of Node
  10.x rather than upgrade to the absolutely latest Node (or even
  the latest LTS track, which is node 12.x)
* Among other things, this patches this vulnerability in npm
  https://blog.npmjs.org/post/189618601100/binary-planting-with-the-npm-cli

### DIFF
--- a/devops/docker/NodeDockerfile
+++ b/devops/docker/NodeDockerfile
@@ -1,8 +1,12 @@
 ARG NODE_VER
+ARG NPM_VER
 FROM node:${NODE_VER}
 
 # Make npm output less verbose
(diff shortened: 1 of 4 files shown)


chigby (approved):
Looks good to me, here. I like how this change makes the versions available as arguments so we can easily change them in the future.

Issue #657, opened 2019-12-16
Expiring cookies in VCR recording


chigby:
In our test suite, we use VCR to record the HTTP responses during a scan of a domain/site. These recordings include cookie data, and cookie expiration dates that are determined when the recording is made. Our tests assert that we properly detect the presence of cookies. However, if the test suite is run against the recordings _after_ the expiration date has passed, then the scanner will (correctly) mark the site as not using cookies. Then any test assertions made about the presence or absence of cookies will fail after that date.

PR #658, opened 2019-12-16
Re-generate VCR file to account for cookie expiration

chigby:
This pull request updates the VCR recording of the "full scan site live" file to account for expired cookies that were causing the test suite to fail.

I'm mostly doing this so that our tests will pass. It's not a general solution to the problem and without some other change, I suspect we will have to do this again in mid-January.

This pull request also fixes a vulnerability found by `npm audit`. The overall theme of this PR is "fix the build."

Refs #657


SaptakS:
This PR also solves the only test failing problems in https://github.com/freedomofpress/securedrop.org/pull/654/files. Though creating new VCR records every month mid doesn't sound best approach to update the dates

PR #661, opened 2020-01-02
Upgrade to Wagtail 2.7 and Django 2.2

### LINKED ISSUE #660: Upgrade to Django 2.2 LTS and Wagtail 2.7 LTS
Refs freedomofpress/fpf-www-projects/issues/43

chigby:
This pull request upgrades the site to Django 2.2 and Wagtail 2.7. See commits for upgrade specific notes. Notable alterations:

* `ReadOnlyPanel`, which was a subclass of wagtail's `EditHandler` had to be modified.
* Something inside of `django.db.models.Value` changed (or possibly `SearchVector`?) and our use of it cause an error, which I've fixed.

I did some cursory testing of the site locally to make sure page editing and rendering still worked. Even so, I recommend that we probably ought to do some checking of basic functionality and content on the sandbox before deploying this.

Fixes #660


forms/utils.py
```diff
@@ -5,7 +5,7 @@
 
 def send_mail(subject, message, recipient_list, from_email=None, **kwargs):
     """
-    Modified clone of wagtail.admin.utils,send_mail that omits the
+    Modified clone of wagtail.admin.utils.send_mail that omits the
```
harrislapiroff: 😅

PR #667, opened 2020-01-21
VCR Enhancements

### LINKED ISSUE #657: Expiring cookies in VCR recording

chigby:
This pull request:

1. Upgrades us to the most recent version of VCR (1.11.1 to 4.0.2)
2. Creates a response header filter function to rewrite the expiration date on cookies received by VCR before putting them in a file. The expiration date is about 12 years in the future.
3. Regenerates, with the new filter, the `full-scan-site-live.yaml` file.
4. Fixes up some problems with the pip-related commands and the `Makefile` that I noticed while performing the package upgrades.

Fixes #657


scanner/tests/test_scanner.py
```diff
@@ -17,6 +20,18 @@
...
+def long_lasting_cookies(response):
+    """modify a HTTP response to extend cookie lifetime"""
+    if 'Set-Cookie' in response['headers']:
+        timestamp = datetime(2032, 10, 31, 13, 14, 15, tzinfo=utc)
+        updated_expiry = re.sub(r'(expires=)([\w, -:]+)', r'\1{}'.format(timestamp.strftime("%a, %d-%b-%y %H:%M:%S %Z")), response['headers']['Set-Cookie'][0])
+        response['headers']['Set-Cookie'] = [updated_expiry]
+    return response
+
+
+mod_vcr = vcr.VCR(before_record_response=long_lasting_cookies)
```
harrislapiroff: This is cool!

PR #680, opened 2020-03-16
Add coverage report in Circle CI 

SaptakS:
Refs: https://github.com/freedomofpress/fpf-www-projects/issues/83

@harrislapiroff this creates a coverage report using the python-coverage package, generates an xml version of report and pushes that to codecov. Then it shows the codecov link generated with the report in the CI. Now as a next step, my question is where and what kind of information should we show?

@conorsch as far as I saw securedrop's codebase, securedrop does a similar thing but uses pytests since securedrop is in flask, rather than django tests like we do in this repo. Hence the coverage report generation command is a bit different. Also, I think right now securedrop shows the coverage report as a button on the repo's README. Is it shown anywhere else? Also in that case, @harrislapiroff should we also show just a codecov badge in the readme?

Correct me if I am wrong about any of the above.

cc: @eloquence


SaptakS:
@conorsch @harrislapiroff I have implemented the coverage without codecov uploading feature. So now the coverage report is shown in the circle CI report itself. I have also added a fail under 80% threshold. This is just an arbitrary threshold set based on the fact that this repository has 85% coverage.

SaptakS:
Coverage report added to the artifacts: https://2903-99369601-gh.circle-artifacts.com/0/htmlcov/index.html

SaptakS:
`modernizr>mkdirp>minimist` this is the path which seems to be failing right now. modernizr depends on mkdirp@0.5.1 and that depends specifically on minimist@0.0.8

SaptakS:
Not sure if there's a way to sort this out cc @chigby @harrislapiroff

chigby:
Related: https://github.com/Modernizr/Modernizr/issues/2524

SaptakS:
@chigby guess we will have to wait then?

chigby:
Yes, it seems that way. I don't think that the vulnerability is a blocker for merging this PR, though.

PR #681, opened 2020-03-24
Inline PDF display for blog pages

### LINKED ISSUE #674: Port inline PDF viewer to securedrop.org
freedom.press has a nice inline PDF stream block type, as can be seen on this post:

https://freedom.press/news/freedom-press-foundation-releases-its-2019-impact-report/

It would be great to port this over to securedrop.org. If we do so, we need to test the widget's functionality (toolbar buttons) across browsers to ensure the CSP is configured correctly.

The use case are documents like whitepapers and audit reports.

chigby:
This pull request adds an `InlinePDFBlock` modeled after the one on freedom.press with additional consideration paid to the experience when the user's browser either (1) does not support displaying PDFs (mobile safari, chrome for android) or (2) has Javascript disabled (Tor Browser in Safest mode). In both cases, a link to download the PDF is displayed instead.

Cross browser compatibility required more machinery than I originally thought it would. There's more explanation in the individual commit messages, but the gist is I added the library `pdfobject` to detect if the browser can render a PDF, and if so, embed it in the page. `pdfobject` handles the fallback in the case of (1) above, and a `<noscript>` tag handles the case of (2). 

I also had to add a shim for wagtail to control the `Content-Disposition` header for these inline pdfs not to be `attachment` (which some browsers will automatically treat as a download rather than a display inline). See https://github.com/wagtail/wagtail/issues/1158

Fixes #674


eloquence:
Happy to help test this on sandbox.

client/common/js/embed-pdf.js
```diff
@@ -0,0 +1,13 @@
+import PDFObject from 'pdfobject';
+
+function embedPdf(el) {
+	var options = {
+		fallbackLink: '<p class="pdfobject-fallback"><strong>PDF Not Displayed</strong>: This browser does not support inline PDFs. Please download the PDF to view it: <a href="[url]">Download PDF</a>.</p>'
```
harrislapiroff: I think it would actually be reasonable for this message not to even indicate that it is a fallback. Maybe just a button with an icon that says "Download <strong>Document Title</strong>" Something like this:

Artboard

Here's an SVG icon for the page

```svg
<?xml version="1.0"?>
<svg xmlns="http://www.w3.org/2000/svg" width="8" height="8" viewBox="0 0 8 8">
    <path d="M0 0v8h7v-4h-4v-4h-3zm4 0v3h3l-3-3z"></path>
</svg>
```

harrislapiroff: Uh, but, like, a reasonable normal button height. Not huge.

common/templates/common/blocks/inline_pdf.html
```diff
@@ -0,0 +1,8 @@
+{% load static common_tags %}
+
+<div class="js-pdf" data-file-uri="{{ self.document|document_view_url}}"></div>
+<noscript>
+	<p class="pdfobject-fallback"><strong>PDF Not Displayed</strong>: This browser does not support inline PDFs. Please download the PDF to view it:
+		<a href="{{ self.document.url }}">Download PDF</a>.
+	</p>
```
harrislapiroff: See above

eloquence:
Needs rebase.

harrislapiroff:
Here's the image as a PNG (in the white space below... white on transparency 😅)

File

chigby:
aw, dang

chigby:
Converted the fallback image to the given PNG, but I kept the old one around for that glorious someday in which we get a tag to handle both at once.

PR #682, opened 2020-03-30
Adds optional square logo for organizations for directory listing

### LINKED ISSUE #563: Add option to DirectoryEntry for a square logo
As initially mentioned [here](https://github.com/freedomofpress/securedrop.org/pull/464#issue-183882956).

- [ ] Add an optional `organization_logo_square` field to `DirectoryEntry`
- [ ] Use that field in the [_instance_table_row.html](https://github.com/freedomofpress/securedrop.org/blob/master/directory/templates/directory/_instance_table_row.html) template, but fall back to `organization_logo` if it is empty
- [ ] (Content change, no code change) Add a field to the [instance submission form](https://securedrop.org/directory/submit/) for an optional square logo

SaptakS:
Fixes #563


directory/templates/directory/_instance_table_row.html
```diff
@@ -2,7 +2,10 @@
 
 <tr class="instance-table__row">
 	<td class="instance-table__icon-column">
-		{% if instance.specific.organization_logo %}
+		{{ instance.specific.organization_logo_square }}
```
chigby: Is this line correct? Seems like it was included here accidentally :sweat_smile:

accounts/forms/directory_management.py
```diff
@@ -60,6 +61,26 @@ def clean_organization_logo(self):
...
+        try:
+            organization_logo_square = WagtailImage.objects.create(
+                title='{} logo'.format(self.cleaned_data['title']),
+                file=organization_logo_square_file,
+            )
+        except Exception:
+            raise ValidationError('That image could not be saved')
+
+        return organization_logo_square
+
```
chigby: So, I'm not completely sure about this, but I kind of suspect that this form (and probably most of the code in the `accounts` app) is not actually being used. Do you know if we're using it? Maybe @harrislapiroff knows? 

The upside is we might not need this addition at all. Even if we are using it somewhere, it looks like this form is somewhat out of date regarding logos anyway, as there are a couple fields (e.g. `organization_logo_homepage`, `organization_logo_is_title`) that are on the `DirectoryEntry` model and not in this form. It's possible this is by design, but my suspicion is it's just because this form is unused. It's possible I'm wrong here, though.

directory/models/entry.py
```diff
@@ -115,6 +115,14 @@ class DirectoryEntry(MetadataPageMixin, Page):
         related_name='+',
     )
 
+    organization_logo_square = models.ForeignKey(
+        'common.CustomImage',
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name='+',
+    )
```
chigby: :+1:

accounts/forms/directory_management.py
harrislapiroff: Yes, that's correct. We're basically not using the `accounts` app.

PR #685, opened 2020-04-06
NPM audit: Remove modernizr and modernizr-loader, update css-loader

chigby:
This pull request removes modernizr from our webpack config and our npm dependencies. It also updates the version of `css-loader` we're using. Both of these packages have security vulnerabilities in their dependencies that are triggering our `npm audit` check.

The modernizr tool is not being used (there is no .modernizrrc file in our repo) and does not need to be installed as part of package.json.  I'm removing it because right now it has a security vulnerability in one of its dependencies:

https://snyk.io/vuln/SNYK-JS-MINIMIST-559764

but in general we shouldn't install software we no longer need (or never needed).


chigby:
I think it's because the wagtail admin includes modernizr as a `<script src="...">` tag.

harrislapiroff:
ahhhhhhhh

PR #688, opened 2020-05-05
Updates to wagtail-2.7.3+

SaptakS:


SaptakS:
Seems markdown and sslyze are also failing.  markdown2 is fine to update. But sslyze we will have to update 2 major releases. Also it seems to be pinned right now to 1.4.1. @harrislapiroff @chigby any particular reason for this?

chigby:
```
-> sslyze, installed 1.4.1, affected <3.0.2, id 38197
Sslyze 3.0.2 improves the check for HTTP security headers by adding support for HTTP redirections.
```

It looks like sslyze 3.0.2, the version not affected by this safety warning, is only available for python 3.7 and 3.8: https://github.com/nabla-c0d3/sslyze/blob/7979e7b3f9535328d796563398733d4495c54016/setup.py#L60

emkll:
In the past sslyze has proven very complicated to update due to pshht and other dependencies. Given this warning is only related to improvement of header checking,  I would recommend we add `--ignore 38197` to the check command, and consider updating pshht/sslzyze at some point in the future once we move to newer Python versions, or if we need the new functionality.

SaptakS:
@emkll thanks!  Will go down that path then.

SaptakS:
Seems markdown2's latest version also has the vulnerability :smile:

PR #689, opened 2020-05-05
Add detection logic for mobile tor browser

### LINKED ISSUE #687: Warning banner Tor Browser Androi
Tor browser for Android gives you the big red warning when that you're not anonymous when visiting pages like securedrop.org/directory. It'd be nice if it was a bit smarter. It currently includes a link to torproject.org, where one could donwload Tor browser for Android. Some people like recursivity, but maybe rewording to use the desktop version would be better. I don't know if removing the banner completely for Tor Browser Android (and maybe other versions?) is a good idea or not.

chigby:
This pull request adds a separate function to detect Mobile Tor Browser and incorporates that into the Tor warning display logic. 

The implementation was copied over from [SecureDrop](https://github.com/freedomofpress/securedrop/blob/ec2220c3c2b9120d029b616d3a07647b175bc6ab/securedrop/static/js/source.js#L78)

Fixes #687


client/tor/js/detect-tor.js
```diff
@@ -17,11 +18,18 @@ const is_likely_tor_browser = function () {
 	)
 }
 
+const is_likely_mobile_tor_browser = function () {
```
SaptakS, line 21: @chigby This definitely works. Just one thought. I understand that Securedrop project uses two different functions because it shows a [different suggestion for desktop and mobile](https://github.com/freedomofpress/securedrop/blob/ec2220c3c2b9120d029b616d3a07647b175bc6ab/securedrop/static/js/source.js#L134-L140). But I think since we are not doing any such thing, maybe we can put it into the same TBB_UA_REGEX regex (which you have already done) and not need this function? Just a thought. cc: @harrislapiroff @eloquence

chigby: I was thinking about that possibility! But I also noticed that for SecureDrop they do not do the window dimensions comparison for the mobile browser, so I thought that combining the two user agent expressions might not be correct? Do you have thoughts on that?

chigby: Maybe that comparison is moot on a mobile browser? I'd believe that.

SaptakS:
@chigby @harrislapiroff I checked with ngrok and seems it still shows the banner in tor browser for android. I will check a few more things and report back

chigby:
@SaptakS I just pushed a version of the changes that splits the mobile detection into a separate function, as SecureDrop does. Let me know if this changes anything when you test it with ngrok (and thanks for doing this :slightly_smiling_face:)

SaptakS:
@chigby seems to be working for me.

PR #695, opened 2020-06-23
Onion location header

### LINKED ISSUE #693: Set Onion-Location header
securedrop.org has a corresponding onion service (v2 for now). We should advertise it via `Onion-Location`, which Tor Browser 9.5 will interpret to show an ".onion available" button.

https://community.torproject.org/onion-services/advanced/onion-location/

Example screenshot:
Example

chigby:
This pull request adds a piece of middleware to set the `Onion-Location` header based on its value in the site's `FooterSettings`. The `https` protocol is added to the URL if it's not present, though it's not a foolproof URL transformation. I mostly looked at what is currently there on the production site, and asked "What's the simplest possible way to transform this into the required format for the onion location header?"

If the actual protocol is `http`, let me know and I can change it.

Fixes #693


eloquence:
We do support HTTPS for securedrop.org (it was only recently re-enabled), so that should work. Onion services provide their own encryption layer, though for v2 services, HTTPS provides useful defense in depth.

conorsch:
The changes here _force_ HTTPS on any Onion URL, which isn't what we want—the v3 URLs won't have HTTPS, for instance, at least not as presently planned. While we could add HTTPS to the v3 URL at some later point, I don't think the middleware should force it, it should honor the [spec](https://community.torproject.org/onion-services/advanced/onion-location/) and permit _either_ http or https:

> The Onion-Location value must be a valid URL with http: or https: protocol and a .onion hostname.

So in terms of validating what's been put into the field by a user of the Wagtail admin, sanity checking that _either_ http or https was prepended, and that a suffix of `.onion` exists, seems like a good idea. We also have the ability to validate that the URL is actually a v2 or v3 onion url via regex, but that seems more complicated than it's worth.

chigby:
Thanks for the insights, everyone. I've reworked the middleware to use the address as-entered, and validation to ensure it's correctly formatted for use in the header. I think this means it will need to be updated from its current value to have the correct "http" or "https" protocol, as right now it doesn't look like it has any. Though this may have to wait until this PR is merged since, as far as I can tell, there's no way currently in the admin UI to actually modify this value.

PR #706, opened 2020-07-20
A couple changes to prepare for k8s deployment

maeve-fpf:
These are all basically the same things we changed on STN. We now have `DJANGO_LOG_CONSOLE` as an env var. DO doesn't have that set, so this could be deployed there.

### DIFF
--- a/devops/docker/NodeDockerfile
+++ b/devops/docker/NodeDockerfile
@@ -10,7 +10,7 @@ RUN apk add --no-cache paxctl python make g++
 RUN paxctl -cm /usr/local/bin/node
 
 ARG USERID
-RUN adduser -D -g "" -u "${USERID}" docker_user || true
+RUN getent passwd "${USERID?USERID must be supplied}" || adduser -D -g "" -u "${USERID}" sdo_node
 USER ${USERID}
 
 CMD npm install && touch .node_complete && npm run start
--- a/devops/docker/NodeDockerfile
+++ b/devops/docker/NodeDockerfile
@@ -1,13 +1,16 @@
-ARG NODE_VER
-ARG NPM_VER
-FROM node:${NODE_VER}
+# sha256 as of 2020-06-30 for node:10-alpine
+FROM node@sha256:bafda8e474ba8973f146cb3fc2d531cf7fb7fec73ef4ca04aa38f26db520a883
 
-# Make npm output less verbose
+# Install npm, making output less verbose
+ARG NPM_VER=6.13.7
 ENV NPM_CONFIG_LOGLEVEL warn
-
-# Upgrade npm to speicifed version
 RUN npm install npm@${NPM_VER} -g
+
+RUN apk add --no-cache paxctl python make g++
+RUN paxctl -cm /usr/local/bin/node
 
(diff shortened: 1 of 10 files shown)


maeve-fpf:
```
gyp verb check python checking for Python executable "python2" in the PATH
gyp verb `which` failed Error: not found: python2
```

:cry:

maeve-fpf:
Added a possibly controversial change: don't uppercase variables that are not settings (https://docs.djangoproject.com/en/3.0/ref/settings). I think this makes it clearer what is there for Django to read and what is our own internal logic to figure out what to set.

securedrop/settings/base.py
```diff
@@ -283,56 +295,75 @@
...
+        'formatter': 'json_out'
+    }
+else:
+    console_log = True
+    log_handler = 'console'
+    log_handlers['console'] = {
+        'level': log_level,
+        'class': 'logging.StreamHandler',
+        'formatter': 'json_out',
+        'stream': sys.stdout,
```
chigby, line 335: This line needs an `import sys` at the top of the file

devops/docker/NodeDockerfile
```diff
@@ -1,13 +1,11 @@
-ARG NODE_VER
-ARG NPM_VER
-FROM node:${NODE_VER}
+# sha256 as of 2020-06-30 for node:10-alpine
+FROM node@sha256:bafda8e474ba8973f146cb3fc2d531cf7fb7fec73ef4ca04aa38f26db520a883
```
chigby, line 2: In changing this to alpine linux, it's possible we lost access to `python` and `make`, which we need to build `node-sass`. It's possible that what is needed here is something akin to what is being done in the production docker file: https://github.com/freedomofpress/securedrop.org/blob/ed18480a3934215685ed0047fc41df102819a1b9/devops/docker/ProdDjangoDockerfile#L11-L13

maeve-fpf: Thanks! Testing these out now.

maeve-fpf:
Made some more updates, ready for re-review! This is rebased on requirements update PR so that it will pass CI.

eloquence:
@maeve-fpf Could you give us an update on the status of this PR? Is it ready for the web team's review at this point?

maeve-fpf:
@eloquence There is a remaining CI timeout issue that I need to address. Will mark as blocked now and remove after rebasing and resolving that.

maeve-fpf:
Marking this as unblocked, but not sure what to do about making it reviewable - would splitting it up into multiple PRs be better?

eloquence:
I defer to Harris & team; it might make sense to do a pair review with an infra team member.

conorsch:
Sounds like a great plan. Since @maeve-fpf wrote the changes, I volunteer to co-review, either with @harrislapiroff or someone else from web.

maeve-fpf:
I've rewound this so that it doesn't include whitenoise, since that can be reviewed separately -- available today to pair on review here if needed.

maeve-fpf:
Also marking as blocked by https://github.com/freedomofpress/securedrop.org/pull/723.

PR #708, opened 2020-07-21
Update CircleCI image and Python 3.5 version

maeve-fpf:
The Python 3.5 container is on 3.5.9; let's match it. In this case, we were already using a CircleCI 16.04 image, but let's update it to current.


maeve-fpf:
NPM check will be fixed by #707.

SaptakS (approved):

PR #709, opened 2020-07-22
Link Error Report Form  on Directory Page

### LINKED ISSUE #648: Link to contact form instead of email address for notifying FPF about onion URL mismatches
Directory entries currently state:

> If you encounter a mismatch, please contact us at securedrop@freedom.press. This could be evidence of an attempted attack, or it could be a simple failure to communicate between an organization running SecureDrop and Freedom of the Press Foundation. In either case, we will investigate the matter and establish a path to resolution. 

We should not be linking to the email address here and instead point to the contact form. That's because prospective sources may be using this address, and email is less secure by default than a form submission via an HTTPS website.

NB: The current https://securedrop.org/help/ form is geared towards news organizations; it may be good to prepare a variant that's not.

prateekj117:
Fixes #648 

For onion address mismatch reporting, we should add a Directory Settings to define an Error Report Page, as now we have a dedicated Error Reporting Page made by @eloquence.


prateekj117:
@harrislapiroff This is ready for review.

directory/templates/directory/_instance_instructions.html
```diff
@@ -22,6 +22,11 @@ <h1 class="instance-instructions__title">{% trans "Planning to submit informatio
 				<a href="mailto:{{ contact_email_escaped }}">{{ contact_email }}</a>
 				(<a href="{{ contact_gpg }}">GPG</a>).
 			{% endblocktrans %}
+		{% elif settings.directory.DirectorySettings.report_error_page %}
```
SaptakS, line 25: Just curious if the report error form should be preferred over email with contact_gpg (given that people might completely ignore the GPG part sometime?) cc: @harrislapiroff @eloquence

prateekj117: Oh yes. I think that is the desired behaviour. I will make that change. Thanks.

directory/models/settings.py
```diff
@@ -66,6 +75,7 @@ class DirectorySettings(BaseSetting):
         MultiFieldPanel([
             FieldPanel('new_instance_alert_group'),
             FieldPanel('contact_email'),
+            PageChooserPanel('report_error_page'),
```
chigby, line 78: One thing that might be nice here, to make it easier for editors to use this panel, is to restrict the allowable pages to FormPages. We have done something similar here:

https://github.com/freedomofpress/securedrop.org/blob/3c9db1164523ab3b95d2eb90936475cb199e755a/directory/models/pages.py#L88

It's just a nice touch that I think is worth using here as well.

prateekj117: @chigby Sure, will do that change.

SaptakS:
I have a feeling (which might be wrong) that the builds are never completing since the PR is made from a fork? Is that possible that maybe since the fork version of CI doesn't have some env vars, it is not completing the build? My theory can be completely wrong.

prateekj117:
@SaptakS You might be right. Have a look at this: https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories. Also, see the topic just underneath it. I think it might be one of these. Can you have a look at the settings, please?

harrislapiroff:
Hmmm unfortunately it seems like our build requires secrets that I'm not sure I should enable passing to forked code. I'll have to work around that by opening my own PRs to merge forked ones for the time being.

harrislapiroff:
Huh looks like I was wrong about that actually.

PR #710, opened 2020-07-26
Checks the hostname and onion address to see if they match before sending header

### LINKED ISSUE #698: Onion available button shows even when onion being used
The `onion-location` header added in #695 causes tor browser to render an Onion available button, which is great! Unfortunately it still renders the button _even if_ the onion is already being used.

We'll chat with Tor about whether this is a UI behavior issue, but, in the meantime, let's fix it for SDO by setting the `onion-location` header conditionally—only if the onion service is not being used. We expect that the onion url will be available to django as the hostname and we can use that to check.

SaptakS:
Might fix #698 

I was unable to check the onion location button locally. For some reason the location button doesn't appear when tried with ngrok, so not sure. I am just checking the request host name with the onion address saved. And if they don't match, I send the location header.

### DIFF
--- a/common/middleware.py
+++ b/common/middleware.py
@@ -7,8 +7,10 @@ class OnionLocationMiddleware(object):
 
     def __call__(self, request):
         response = self.get_response(request)
-
         footer_settings = FooterSettings.for_site(request.site)
-        if footer_settings.securedrop_onion_address:
-            response['Onion-Location'] = footer_settings.securedrop_onion_address
+        if (
+            footer_settings.securedrop_onion_address
+            and request.get_host() not in footer_settings.securedrop_onion_address
+        ):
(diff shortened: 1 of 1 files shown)


eloquence:
Since there's no pressing need for this change (it's already enforced at the infra level), we've agreed to wait with formal review/merge of this PR until we have a staging environment for the SecureDrop.org .onion service, which seems achievable within the next month as part of the ongoing k8s transition. Moving to backlog for now.

(Note that a Tor Browser upstream fix may ultimately make this change unnecessary, although we may still end up keeping it around.)

SaptakS:
Closing this PR given [the issue](https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/40019) got fixed in upstream

PR #712, opened 2020-08-04
Link error report page

### LINKED ISSUE #648: Link to contact form instead of email address for notifying FPF about onion URL mismatches

harrislapiroff:
Duplicate of #709 from inside the repo for CI purposes.

> Fixes #648
>
> For onion address mismatch reporting, we should add a Directory Settings to define an Error Report Page, as now we have a dedicated Error Reporting Page made by @eloquence.


harrislapiroff:
Never mind this wasn't necessary.

PR #715, opened 2020-08-06
Display a more clear message when Instance is down for Maintenance

### LINKED ISSUE #700: Add custom delisted template for "extended downtime" case
The options for delisting a SecureDrop instance (in the settings for a directory entry) distinguish various reasons, such as "Non-200 status response" and "Extended downtime". However, all of those reasons appear to currently result in the same display on the directory entry, a page which says:

> Freedom of the Press Foundation routinely verifies that SecureDrop instances are up and running and follow best security practices. We have detected issues with this instance, and have temporarily removed it from the directory.

This language is problematic for a case where an instance has simply been down for maintenance for a while. That's why I don't use delisting at all right now in those cases, and instead unpublish them entirely -- which is not great, because it means folks will just get a 404 if they try to visit an entry they've gone to before.

I would suggest that we set up a template for that case, which would say something like:

> This instance is experiencing an extended service outage. We have temporarily removed it from the directory.

We can continue to use the other template for all other "delisted" reasons that are currently supported.

prateekj117:
Fixes #700 

The old instance alert message language was problematic and misguiding when the instance is simply down for maintenance. So it is better if we add a separate message if the instance is simply down for maintenance.


prateekj117:
@eloquence Ready for review.

directory/templates/directory/_instance_maintenance_alert.html
```diff
@@ -0,0 +1,13 @@
...
+	{% include "common/_svg.html" with class="instance-alert__icon" svg="images/instance-status/Pending.svg" img="images/instance-status/Pending.png" alt="Verification Pending Icon" %}
+	<h2 class="instance-alert__title">
+		{% trans "Instance Under Maintenance" %}
+	</h2>
+	<div class="instance-alert__content">
+		{% blocktrans %}
+			This instance is experiencing an extended service outage. We have temporarily removed it from the directory.
+		{% endblocktrans %}
+	</div>
+</div>
```
SaptakS, line 13: Nitpicking, please add a line at the end of file.

SaptakS (changes requested):
@prateekj117 looks almost ready to be merged! The migration file is missing from the PR. Please add the corresponding migration file related to the change in [DELISTED_REASONS](https://github.com/freedomofpress/securedrop.org/pull/715/files#diff-60b040892be58a6c734077904ce23effR170)

prateekj117:
@SaptakS Ready for another review.

SaptakS (approved):
Looks good to me

prateekj117:
@harrislapiroff Ready for review.

Issue #701, opened 2020-07-10
"Pls use Tor" banner interferes with usability on mobile


eloquence:
The "You are not anonymous while using this browser" takes up the whole screen on mobile browsers, interfering with basic site usability. In Firefox/Tor mobile versions, it also renders incorrectly for me, with the "X" to close not visible without scrolling horizontally, making the site very difficult to use.

Moreover, the guidance is misleading on mobile: If you download the mobile version of Tor Browser, you will still get the banner (we don't recommend using mobile versions of Tor for actual SecureDrop submissions).

We should come up with more appropriate security guidance for mobile users, and consider a design that's less disruptive of other site uses.

prateekj117:
@eloquence Will look into this.

ninavizz:
The messaging for mobile should also be unique, as there are only 3rd party mobile apps for iPhones which may or may not be safe. Android has an official app from Google Play or FDroid.

harrislapiroff:
Sounds like there's a couple problems to fix here:

1. a rendering issue where the banner is impossible to dismiss on certain mobile browsers
2. a more conceptual messaging issue where we should be offering a different message to mobile users (maybe even different messages to android vs iphone users)

Thanks for looking into 1, @prateekj117! Keep me posted on what you find; we'd like to resolve this fairly urgently. For 2, I'm going to open a separate ticket.

eloquence:
> The messaging for mobile should also be unique

Agreed; we generally want to direct folks to desktop if they want to submit tips. I would suggest the following language:

> Have a document to share?
> 
> Follow these steps to protect your anonymity:
> 
> 1. Choose a safe location, not a work computer.
> 2. Download the desktop version of Tor Browser from https://www.torproject.org/
> 3. Launch Tor Browser and return to this site. You are are now anonymous online.

This is an oversimplification but better than what we have now. It should be shown for mobile user agents, including mobile versions of Tor Browser.

ninavizz:
@eloquence I want to be sensitive to the reality that many disadvantaged folks don't have laptops; nor, access to them. It feels restrictive and problematic to not acknowledge Tor Mobile options in any capacity.

eloquence:
> @eloquence I want to be sensitive to the reality that many disadvantaged folks don't have laptops; nor, access to them. It feels restrictive and problematic to not acknowledge Tor Mobile options in any capacity.

[...]

PR #717, opened 2020-08-11
Special warning for mobile tor browser

### LINKED ISSUE #703: Tor browser modal should show a different message on mobile
See discussion under #701

chigby:
This pull request adds the appropriate title/subtitle/body settings fields to the Tor alert settings for creating a new message alert box. This content will be shown to users of the site who we detect are browsing with the Mobile Tor browser.

Fixes #703


client/tor/js/detect-tor.js
```diff
@@ -65,6 +65,33 @@ document.addEventListener("DOMContentLoaded", () => {
...
+			torWarning.classList.add('tor-warning--hidden')
+			// hides warning for screen readers
+			torWarning.setAttribute('aria-hidden', 'true')
+			body.classList.remove('no-scroll')
+			if(instances) {
+				instances.classList.remove('instances--tor-warning')
+			}
+			sessionStorage.setItem('torWarningDismissed', '1')
+		})
+	} else if (is_likely_mobile_tor_browser()) {
```
SaptakS, line 78: I thought the discussion in #701 was to show the mobile warning in all the mobile browsers, irrespective of whether it's a tor based browser or not. Won't this check make it show the mobile warning only in mobile tor browser? cc @eloquence 

Also, showing with javascript also has the problem that in the safest security setting, when JS is disabled, this warning won't appear as[ I mentioned here](https://github.com/freedomofpress/securedrop.org/pull/718#issue-466686539)

chigby: All right, thanks for pointing this out. I have updated the code to check for all mobile browsers.

PR #718, opened 2020-08-12
Refactors tor warning for mobile devices

SaptakS:
Refs (and hopefully fixes) #701 

Few things that are addressed in this PR:
- The close button should now be easily visible in both tor browser and firefox in mobile devices.
- I have added a field called `mobile_subtitle` in the TorAlertSettings which by default is set to the message `for security reasons we discourage use of mobile browser` as discussed in [the comments](https://github.com/freedomofpress/securedrop.org/issues/701#issuecomment-659668650). Anyone can go and update the copy from the admin settings
- Removed the check of detecting mobile tor browsers to remove the warning (Since we decided to show in all mobile browsers, tor or not) [Haven't deleted the actual function yet]

Few things I think might be worth considering:
- Right now, it will show the same set of instructions in the tor warning body as shown in desktops. In case we want to show a different set of instructions, I can add another richtext field to be shown in mobile.
- Also, if mobile tor browser has the safest setting, the warning is not shown. This is because we have a hidden by default approach. Which means the warning is hidden by default, and using javascript we remove the hidden class. We could go to opposite approach (especially given that we show warning in most browsers) but the problem then would be the warning wont be hidden in highest security level of tor desktop browser.
- There is actually a second way of handling the above mentioned issue. And that is by using media breakpoints. So if it is a mobile screen, we don't hide even if the hidden class is present. cc: @harrislapiroff 

@eloquence @ninavizz Let me know what you think.


SaptakS:
Oops. Seems @chigby has already opened a [related PR](https://github.com/freedomofpress/securedrop.org/pull/717) for adding the custom tor warning message. Closing this.

PR #719, opened 2020-08-12
Adds styling to improve mobile tor banner

### LINKED ISSUE #701: "Pls use Tor" banner interferes with usability on mobile

SaptakS:
Fixes #701

### DIFF
--- a/client/common/sass/_tor.sass
+++ b/client/common/sass/_tor.sass
@@ -32,6 +32,7 @@
 		margin: 0
 		display: block
 		overflow: scroll
+		max-width: 90vw
 
 
 // This is an override of .tor-warning


chigby:
So, I did some testing of this and the code seems fine as it is, but I want to mention that I was not able to reproduce the original issue of a horizontal scrollbar hiding the close link on the Tor warning (on securedrop.org) using mobile Firefox, desktop Firefox & Tor using various small-sized screen widths.

SaptakS:
Interesting. I was able to reproduce it perfectly. The banner was overflowing on the right side. Not on chrome, but on mobile firefox and tor (though not on responsive mode when doing via desktop)

chigby:
It's possible it's just the phone I used for testing (iphone 8 plus) is a slightly different size from others. I'll consider it fixed it you are seeing it fixed.

SaptakS:
I did test in tor and firefox via ngrok and looked fixed to me.

chigby:
Sounds good to me, in that case.

SaptakS:
@harrislapiroff SDO remains unaffected by setuptools

PR #720, opened 2020-08-18
Add .idea to gitignore

### LINKED ISSUE #705: Add .idea to gitignore
We should also add `.idea` to `.gitignore` as it is used by Jetbrains IDE for storing project-specific settings of a local environment.

prateekj117:
Fixes #705 

It is used by Jetbrains IDE for project-specific settings and should be ignored


chigby (approved):
Looks good to me. Nice to get better support for different workflows into our repo.

PR #725, opened 2020-08-25
Add Whitenoise

maeve-fpf:
This adds Whitenoise to the app, to support serving static files. Not very useful on its own, but will be a base for adding Google Storage support later - there will be no nginx to serve static files on the pod that gets deployed in k8s.

Actually adding the middleware is just one line - the bulk of this is making it so that we can run `./manage.py collectstatic` in ProdDjangoDockerfile. Having DEBUG turned on in base.py makes that not work, so this changes the default arrangment to:

- base.py: debug is off
- dev env (including running manage.py separately): debug is on unless you set DJANGO_DISABLE_DEBUG env var
- prod env (including prod-docker-compose): debug stays off unless you set DJANGO_DEBUG_PROD

and the warning comment about not enabling debug in real prod is moved to production.py. I think that's pretty safe, but if you object to the env var names, we could change that.

Also fix it so that dev.py doesn't assume DJANGO_DB_HOST is set - just use pg if DJANGO_DB_HOST is set and sqlite3 if not.


securedrop/settings/production.py
```diff
@@ -42,6 +46,8 @@
 STATIC_ROOT = os.environ['DJANGO_STATIC_ROOT']
 MEDIA_ROOT = os.environ['DJANGO_MEDIA_ROOT']
 
+MIDDLEWARE.append('whitenoise.middleware.WhiteNoiseMiddleware')  # noqa: F405
```
chigby, line 49: So, I'm slightly worried that the particularities of whitenoise make the ordering-of-middleware situation more complicated than we'd like. The [documentation](http://whitenoise.evans.io/en/stable/django.html#enable-whitenoise) says that whitenoise is supposed to be placed "after Django's SecurityMiddleware and before all other middleware." 

I poked around a little with this locally by running `docker-compose -f prod-docker-compose.yaml up` and with the `MIDDLEWARE.append` logic above, the site was not serving any static files. When I changed it to 

```
MIDDLEWARE.insert(7, 'whitenoise.middleware.WhiteNoiseMiddleware')
```

(where index 7 is just after where the security middleware is in the base list of middlewares), then it did work and static files started showing up locally. But it's also quite possible that some other peculiarity with my local setup could be confusing things.

maeve-fpf: Would an approach like this work? https://github.com/freedomofpress/securedrop.org/commit/47c9656838d082c63d1b191a50be0a12bfaa55b2

While we know it's going to be in there somewhere, avoid possibly raising an exception in case it's not, and still work if we move the other middleware around.

chigby: I think that's a great solution, though I think it might need one modification, which is to have `position` set to the index of security middleware **plus 1**, to put it one slot after the security middleware (sorry if my original wording was not entirely precise in this matter). If we don't add one it will put the white noise middleware before the security middleware.

maeve-fpf: Right, thank you! Okay, I've amended that and added it to this branch. Confirming static files are served from localhost:8000 in my prod-docker-compose environment.

maeve-fpf: Oh, and in case you hit any permissions problems, it's probably old volumes; I had to do `docker-compose -f prod-docker-compose.yaml down --volumes` before starting this.

maeve-fpf: Out of curiosity, is it possibly to set `# noqa: F405` for a whole group of lines? I had to add it to each reference to MIDDLEWARE to make flake8 pass.

chigby: Hmmm.. I'm not seeing anything about that here: https://flake8.pycqa.org/en/latest/user/violations.html, which is a shame

securedrop/settings/production.py
```diff
@@ -42,6 +46,14 @@
 STATIC_ROOT = os.environ['DJANGO_STATIC_ROOT']
 MEDIA_ROOT = os.environ['DJANGO_MEDIA_ROOT']
 
+if 'django.middleware.security.SecurityMiddleware' in MIDDLEWARE:  # noqa: F405
+    # see http://whitenoise.evans.io/en/stable/django.html#enable-whitenoise
+    position = MIDDLEWARE.index('django.middleware.security.SecurityMiddleware') + 1  # noqa: F405
+else:
+    position = 0
+
+MIDDLEWARE.insert(position, 'whitenoise.middleware.WhiteNoiseMiddleware')  # noqa: F405
```
conorsch, line 55: These changes will add whitenoise to _all_ instances of the SDO webapp, even when not hosted on containers. Are there side-effects to adding it to the non-containerized deployments? If not, then perhaps we can omit the insertion logic and just include whitenoise in its proper place (i.e. immediately after security) in the base settings. The whitenoise docs provide guidance on using whitenoise in dev https://whitenoise.evans.io/en/stable/django.html#using-whitenoise-in-development and I don't see any red flags there.

securedrop/settings/production.py
conorsch: @maeve-fpf You can ignore per-file https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-per-file-ignores For the settings files, sounds appropriate for F405 specifically.

maeve-fpf: And that's what we do in STN, I see: the code is

```
if bool(os.environ.get('DJANGO_WHITENOISE', False)):
    security_middle_position = [p for p, v in enumerate(MIDDLEWARE)
                                if "security.SecurityMiddleware" in v]
    MIDDLEWARE.insert(security_middle_position[0]+1,
                      'whitenoise.middleware.WhiteNoiseMiddleware'
                      )
```

and then `.flake8` contains

```
per-file-ignores =
    securethenews/settings/*:F405
```

So that's an option! It does seem like this is something that would come up a whole lot in settings files.

securedrop/settings/production.py
maeve-fpf: From my earlier discussion with @harris, IIRC, it seems like they are two different ways to approach getting acceptable performance (writing them somewhere else for a frontend webserver to read, or using optimized middleware). I think it would be fine to turn it on in dev, if web team people agree! We could also add the same environment variable switch that the STN code above uses.

chigby: FWIW, I don't really have an opinion one way or the other about whitenoise in the dev environment.

maeve-fpf: OK, I added the option to enable it with DJANGO_WHITENOISE or leave it off, so you can opt in if you want. It's in base.py, so it just splits up the big list in the code and does an append in the middle.

securedrop/settings/dev.py
```diff
@@ -9,6 +9,10 @@
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = '-hf!6+rx$-55pyf6tekfkers#7cfn-_d#4f6*vnr-+vz82lqz_'
 
+DEBUG = True
+if os.environ.get('DJANGO_DISABLE_DEBUG'):
+    DEBUG = False
```
chigby, line 14: So one final thing I noticed, which doesn't really have to do with whitenoise but is possibly relevant, is when I tried running the dev server locally with `DJANGO_DISABLE_DEV=1`, I got this message:

```
CommandError: You must set settings.ALLOWED_HOSTS if DEBUG is False.
securedroporg_django_1 exited with code 1
```

And I got around it by adding this line inside the `if` block here:

```
ALLOWED_HOSTS = ['localhost', '127.0.0.1']
```

I dunno if it's a huge concern, but it is something that could be useful.

maeve-fpf: That could be useful! So actually, in `prod-docker-compose.yaml`, we have

```
DJANGO_ALLOWED_HOSTS: app localhost
```

in the environment section, and then some code in production.py looks at that. If it's useful to run the server with debug off in dev, we could move that bit:

```
ALLOWED_HOSTS = os.environ.get('DJANGO_ALLOWED_HOSTS').split(' ')
```

into base. Maybe you just want `*` in dev, I don't know.

PR #726, opened 2020-08-26
Add Google Storage media support

maeve-fpf:
Adds support for storing media/documents in a GS bucket. This is enabled in the k8s staging environment with these environment variables:

  - `GS_BUCKET_NAME=fpf-wagtail-media-sdo-staging` (this is the one that enables the support... it's a Terraform created bucket)
  - `GS_CREDENTIALS=/secrets/credentials.json` (volume for this attached for you by k8s config)
  - `DJANGO_CSP_IMG_HOSTS=storage.googleapis.com` (this is where the bucket lives)
  - `DJANGO_CSP_OBJ_HOSTS=storage.googleapis.com` (ditto; we need it for embedded PDFs but I didn't want to put that in default-src)

some new vars are left as the default (`GS_MEDIA_PATH`, because we don't want to change those URLs from /media/, `GS_PROJECT_ID`, because the bucket is in the same project as the cluster. We also don't use GS for static files at all, yet; that's what Whitenoise is for, and that's probably fine for our scale. But it's a possibility if at some future date we want to upload to GS at build time).

This turned up a bug: there's a part in the document view that tries to remove a `Content-Disposition` header, which I guess was being added in the old setup and it was bothersome to be prompted to download instead of being able to view e.g. a PDF. This doesn't get set in this setup, and the code didn't check before trying to read the value, so there was a KeyError. It should handle that case now.

You can see this in action at: https://staging.securedrop.org/news/piloting-securedrop-workstation-qubes-os/ (a blog post with both an image and an embedded PDF; they should just work :tm: . You can view the media/docs in Wagtail admin although I didn't try modifying any of them).

The part I don't really like is that, on the deployment side, I have nothing better to add to the CSP than `storage.googleapis.com`. That's the host for every GS bucket in the world, which means it is not at all safe to put into production (I feel safe putting it on staging since only FPF org members can access that). We will need to set up a custom endpoint, possibly requiring provisioning a LB over in GCE land or setting up Cloudflare to front the files to the public.

With that done, I'm open to either leaving that endpoint name as something we can configure in the deployment, or if you'd feel safer having it hardcoded in the settings, reverting them to static tuples.

### DIFF
--- a/common/storage.py
+++ b/common/storage.py
@@ -1,0 +1,10 @@
+from django.conf import settings
+from storages.backends.gcloud import GoogleCloudStorage
+
+
+class StaticStorage(GoogleCloudStorage):
+    location = settings.GS_STATIC_PATH
+
+
+class MediaStorage(GoogleCloudStorage):
+    location = settings.GS_MEDIA_PATH
--- a/common/views.py
+++ b/common/views.py
@@ -31,10 +31,11 @@ def view_document(request, document_id, document_filename):
     response = serve.serve(request, document_id, document_filename)
 
     # Remove "attachment" from response's Content-Disposition
-    contdisp = response['Content-Disposition']
-    response['Content-Disposition'] = "; ".join(
-        [x for x in contdisp.split("; ") if x != "attachment"]
-    )
+    if 'Content-Disposition' in response:
+        contdisp = response['Content-Disposition']
+        response['Content-Disposition'] = "; ".join(
+            [x for x in contdisp.split("; ") if x != "attachment"]
+        )
 
     # Force content-type for pdf files
     if document_filename.split('.')[-1] == 'pdf':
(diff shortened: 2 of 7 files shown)


chigby (commented):
All of this is looking good to me. The staging site demo of the images and embedded PDF is great (thanks for fixing the content disposition bug, too!).

Do you think we should merge this in now, or wait until the CSP situation has a better resolution via adding the custom endpoint?

maeve-fpf:
I am OK with merging this now, as staging is behind GitHub auth. When we have a custom endpoint solution, it can be validated in staging first.

chigby (approved):
Approving based on discussion

chigby:
Sounds good. I can merge it after it's rebased.

maeve-fpf:
Rebased and confirmed base is set to `develop`. Thanks!

PR #729, opened 2020-09-01
Check for http2 support during scan

### LINKED ISSUE #684: Consider checking HTTP version in landing page scanner
[As discussed in internal comms]

It could be a good idea to check whether an instance's landing page is using HTTP/1.1 or HTTP/2 as part of our landing page scan. 

There are a few benefits to HTTP/2, in particular, the enforcement of modern cipher suites if HTTPS is used (TLS 1.2 at minimum) [1] and a blacklist of insecure cipher suites [2]. It is not a requirement, but by tracking this, we may be able to support specific orgs with an eventual transition.

[1] https://http2.github.io/http2-spec/#TLSUsage
[2] https://http2.github.io/http2-spec/#BadCipherSuites

chigby:
This pull request updates the scanner with support for checking the support of the HTTP/2 protocol and storing the result as a true/false field. If, when scanning a site, we are able to make a connection to the "canonical" https domain, it will check that connection for support of the protocol and record that information in the scan result.

Currently, the information is not being used for any other site function, but it is viewable in the admin.

I left some notes about the implementation of the check in the commit message.

Fixes #684


SaptakS (changes requested):
@chigby I think you need to rebase and fix the migrations given that it doesn't point to the latest migration.

chigby:
@SaptakS Migrations are fixed now, thanks for looking into that

harrislapiroff:
@SaptakS will you do a favor and rereview now that migrations are fixed? Thanks!

SaptakS (approved):
LGTM

harrislapiroff (approved):

PR #732, opened 2020-09-28
Removes unrequired safety ignores

SaptakS:
It removes the unnecessary safety ignores right now. Still needed to keep the safety ignore related to sslyze because pshtt still has an older dependency

Refs: https://github.com/freedomofpress/fpf-www-projects/issues/122


chigby (approved):
The checks still pass with the removed ignores, so it seems good to me!

PR #733, opened 2020-09-28
Onion location paths

### LINKED ISSUE #724: Onion location header redirects to homepage
The "Onion location header" introduced in https://github.com/freedomofpress/securedrop.org/pull/695/ redirects the user to the onion version of the securedrop website. 

The user is being redirected to the homepage instead of to the page where they were, which is quite annoying.

A lot of websites had this problem. See how they fixed it on the [Qubes website](https://github.com/QubesOS/qubesos.github.io/commit/216afd0fa12528fed3b6ef84dfcd59823bc18d30) for inspiration.

chigby:
This pull request fixes the problem with the `Onion-Location` header always referring to the home page, rather than the page path requested by each individual page.

Fixes #724


SaptakS (commented):
So right now if the `footer_settings.securedrop_onion_address` is set to an onion address without `https://` in the beginning (which is the case in createdevdata), the `urljoin()` function technically just replaces the onion address with just the request.path. Hence only the path is returned in that case. I am not sure in which format onion address is saved in prod. Maybe can add a check of some sort to append `https://` in the beginning?

chigby:
I think there is a validator on the model itself:

https://github.com/freedomofpress/securedrop.org/blob/e67c78c9435e437cc7659b1e787fb2e9e9ee67b2/common/models/settings.py#L43-L52

Though I also do see the benefit of ensuring we don't do the wrong thing in this middleware either, regardless of the overall format.

chigby:
Maybe create dev data is what needs to be changed. Hm.

SaptakS:
> Maybe create dev data is what needs to be changed. Hm.

Possible. Hence marked it as a comment and not request changes. I think given the validator, probably safe to just update create dev data

chigby:
Added! Thanks for the review.

SaptakS (approved):

harrislapiroff:
Given that @maeve-fpf is deploying the k8s powered version of this site tomorrow which will include nginx managed onion-location headers, this PR is no longer necessary (I didn't get my review together in time, whoops). Closing without merging.

PR #736, opened 2020-10-06
Use custom endpoint for Google Storage

maeve-fpf:
Closes https://github.com/freedomofpress/infrastructure/issues/2885

We now have the bucket for media files in GS hosted on a custom domain. This updates the django-storages config to use those URLs. The deployment already adds the correct domain to the CSP.

Setting doc: https://django-storages.readthedocs.io/en/latest/backends/gcloud.html

### DIFF
--- a/common/storage.py
+++ b/common/storage.py
@@ -4,7 +4,9 @@ from storages.backends.gcloud import GoogleCloudStorage
 
 class StaticStorage(GoogleCloudStorage):
     location = settings.GS_STATIC_PATH
+    default_acl = 'publicRead'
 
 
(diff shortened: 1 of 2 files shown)


maeve-fpf:
A note here: the `default_acl` is not changing anything about the files; they are public, but the part of the django-storages code that builds the URL (which works slightly differently if a custom endpoint is set) intentionally doesn't load the object from the bucket for performance reasons, so it needs to know to generate a bare URL instead of a signed one.

chigby (approved):
Having looked at the docs a little, it seems like this PR is doing the right thing here. I'm approving it.

PR #738, opened 2020-10-08
If required env vars are not supplied to django-start.sh, fail

maeve-fpf:
While debugging something, I tried to start the app without `DJANGO_APP_NAME` exported and ran into a cryptic error about importing `.wsgi`. Currently, if that variable is unset, the script will run `gunicorn` but pass in an incorrect module name (which I think `__import__` tries to find as a relative import).

To print an error instead of trying to start the app, this uses the shell feature of failing a command if the interpolated variable is missing (which will cause the script to exit since `set -e` is in use).

### DIFF
--- a/devops/docker/django-start.sh
+++ b/devops/docker/django-start.sh
@@ -16,7 +16,7 @@ wait_for_node() {
 
 wait_for_postgres() {
     echo "Waiting for postgres to start..."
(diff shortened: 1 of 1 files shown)


chigby (approved):
This looks good to me. I tried it out locally and, indeed, without those variables defined, the error message is much more comprehensible.

PR #743, opened 2020-10-13
Remove OnionLocationMiddleware

### LINKED ISSUE #741: Remove onion-location header
We're going to serve this via nginx instead of the application. Leaving it in causes us to serve two `onion-location` headers under our k8s setup, which confuses tor browser.

chigby:
Fixes #741


SaptakS (approved):

harrislapiroff (approved):
So elegant!

PR #745, opened 2020-10-14
Remove onion address field from footer_settings and use deployment's onion service name

### LINKED ISSUE #742: Remove onion address setting from database
Instead of pulling the onion address from the database, we should pull it from settings. Check with @maeve-fpf for the name of the env var that stores the address

SaptakS:
Fixes #742 

Pulls the onion address from the settings, via the environment variable instead of db


eloquence:
@SaptakS Note that the footer currently (prior this PR) does not allocate enough space for a v3 address, so prod layout is currently borked (note the overlapping text on the left):

Screenshot from 2020-10-14 16-48-19

Not filing a dedicated issue in case this is already fixed in your PR, or can be fixed as part of it.

SaptakS:
@eloquence I will handle that in this PR itself. Thanks for pointing this out!

SaptakS:
This is how it looks after the changes:

image

In smaller screens:

harrislapiroff (approved):
Looks good!

PR #746, opened 2020-10-14
Autoplaying media file blocks

### LINKED ISSUE #744: Add wagtailmedia
We should take an approach similar to https://github.com/freedomofpress/freedom.press/pull/617 and make a media block that can either be autoplaying silent or a press-to-play video

chigby:
Fixes #744 

This pull request adds `wagtailmedia` and creates the relevant blocks (with styles and templates) for posting video files onto a page. Then I added this block to all appropriate `StreamField`s I could find. 

Note that I've changed the label on the old "Video" block to differentiate it from this new block.


common/blocks.py
```diff
@@ -61,7 +62,7 @@ class AlignedEmbedBlock(blocks.StructBlock):
     class Meta:
         template = 'common/blocks/aligned_embed.html'
         icon = 'media'
-        label = 'Video'
+        label = 'Embed Video Link'
```
harrislapiroff, line 65: Honestly feel like we should rename the subblock and just call this an embed block, but that sounds too complicated to do right now.

common/templates/common/blocks/video.html
```diff
@@ -0,0 +1,7 @@
+<video
+	class="page-video"
+	{% if self.muted_autoplay_and_loop %}autoplay muted loop{% else %}controls{% endif %}
```
harrislapiroff: I think we also want `playsinline` here as seen here: https://github.com/freedomofpress/freedom.press/pull/767/files

```suggestion
	{% if self.muted_autoplay_and_loop %}autoplay muted loop playsinline{% else %}controls{% endif %}
```

harrislapiroff (changes requested):
Gonna make this change myself

harrislapiroff (approved):

harrislapiroff (approved):
With my change, this looks great!

PR #752, opened 2020-10-15
Add a CSP_MEDIA_SRC setting that pulls from the env when needed

harrislapiroff:


maeve-fpf (approved):
Looks good, and given that videos may just be replacements for animated GIFs, no problem with reusing the existing variable. In the future perhaps it could be consolidated down to one var.

PR #755, opened 2020-10-21
Update to Python 3.7

### LINKED ISSUE #568: Upgrade to python 3.7
SSLyze (used by pshht) has a hard requirement for Python >=3.6. Currently we use Python 3.5.3 for SecureTheNews. Because these dependencies are highly coupled, it will make updating them difficult as time goes by.

We are also far behind on pshtt version https://github.com/freedomofpress/securedrop.org/blob/master/dev-requirements.txt#L56 (latest as of this writing is 0.5.2)

Recently there have been several security fixes for SSL-related libraries. While they do not directly affect SecureTheNews, they will make upgrades more complex should a rapid version be required.

Let's move to Python 3.6 to ensure we can timely update dependencies, should a security vulnerability require rapid patching.

chigby:
This pull request updates the python version used in our docker containers from 3.5 to 3.7.9.

I've also included some changes to the Makefile to bring it more into line with our similar repositories. This made it easier for me to do the package updates necessary for bringing everything up to date with python 3.7, and I figured it would be nice to commit these updates here, too.

Fixes #568


harrislapiroff (approved):
This looks good to me! Let's get it on staging, but save deploy for early next week.

PR #762, opened 2020-10-27
Social media links

### LINKED ISSUE #696: Add Twitter/Mastodon links to footer
We currently don't showcase our social links anywhere on securedrop.org. I suggest we add the following as icon links in the footer:
https://twitter.com/securedrop
https://fosstodon.org/@securedrop (new)

This is the recommended icon for Mastodon account: https://fontawesome.com/icons/mastodon

If we use `rel="me"` for the link, Mastodon will render our account as _verified_.  (See docs [here](https://docs.joinmastodon.org/user/profile/)).

chigby:
Fixes #696 

This pull request adds settings for mastodon and twitter links, and displays them in the footer if supplied. Follows the basic design described in the issue, though with the Onion URL below the links due to its length.

image


eloquence:
Can we add support for `rel="me"` for the Mastodon link? That way our account will show up as verified. See docs here: https://docs.joinmastodon.org/user/profile/#verification

chigby:
Just added it. Thanks for the reminder!

common/templates/common/_footer.html
```diff
@@ -66,6 +66,18 @@ <h3 class="sr-only">Footer navigation</h3>
...
 								</div>
 							{% endif %}
+							{% if footer.twitter_url or footer.mastodon_url %}
+								<div class="footer-main__social-media">
+									{% if footer.mastodon_url %}
+										<a href="{{ footer.mastodon_url }}" rel="me">{% include "common/_svg.html" with class="footer-main__social-icon" svg="images/mastodon.svg" img="images/mastodon.png" alt="Mastodon Profile" %}</a>
+										<!-- Font Awesome Free 5.15.1 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free (Icons: CC BY 4.0, Fonts: SIL OFL 1.1, Code: MIT License) -->
+									{% endif %}
+									{% if footer.twitter_url %}
+										<a href="{{ footer.twitter_url }}">{% include "common/_svg.html" with class="footer-main__social-icon" svg="images/twitter.svg" img="images/twitter.png" alt="Twitter Profile" %}</a>
```
SaptakS: Given that these are external links, do we want to add `target="_blank" rel="noopener noreferrer"` ? Though I noticed that in directory listings, we don't really use target=_blank, so maybe not needed. What do you think?

client/common/sass/_footer.sass
```diff
@@ -116,6 +116,14 @@ $list-bottom-padding: 2rem
 		display: block
 		width: 100%
 
+.footer-main__social-icon
+	width: 64px
+	padding-right: 1em
```
SaptakS: Wondering if giving the width based on `rem` makes more sense. Also, probably using `margin-right` makes more sense here? Padding also seems to change the position of the securedrop logo in footer and also reduces the actual width.

SaptakS (commented):
To my eyes, the icons look a bit big compared to all the texts in the area. Especially in a laptop screen. Though I see that it is mostly same as per the design in the issue. So I will leave the final decision to @harrislapiroff .

client/common/sass/_footer.sass
chigby: This makes a lot of sense. I think I started off with pixels just to get a sense of these icons and their placement and never went back and changed it :sweat_smile:

common/templates/common/_footer.html
chigby: Even if we're not using it widely, I think it's probably a good practice to be doing that, so we might as well start here. The only reason maybe not to do that is if Mastodon treats that other rel metadata differently, but from looking at the docs I think it's fine.

SaptakS (approved):
Looks good to me.

harrislapiroff (approved):
Looks good!

PR #764, opened 2020-10-28
Scan results in directory API

### LINKED ISSUE #734: Add landing page scan results to API
In #527 we added a `manage.py` command that exports the landing page scan results in CSV format, which is handy for quickly determining which SecureDrops have significant landing page issues (e.g., use of Google Analytics). This command is run as a cron job and the result is published to http://securedrop.org/media/documents/scanresults.csv (we've never announced that URL, but it is referenced e.g. in #532). 

This is adding to the technical complexity of SecureDrop.org, and we may temporarily lose this CSV as part of the migration to k8s. If possible, it would be good to make the latest scan result available via the directory API, as we do for SecureTheNews.

chigby:
This pull request adds a `latest_scan` field to entries in the directory API. This field contains the complete information for the most recent scan result for that directory. I've also included some minor updates to some related test factories.

To test this, you can load the url http://localhost:8000/api/v1/directory/ in a browser or with a tool like `curl`. 

image

Fixes #734

### DIFF
--- a/directory/api/serializers.py
+++ b/directory/api/serializers.py
@@ -1,7 +1,7 @@
 from rest_framework import serializers
 from wagtail.images.api.fields import ImageRenditionField
 
-from directory.models.entry import DirectoryEntry
+from directory.models.entry import DirectoryEntry, ScanResult
 
 
 class DirectoryEntrySerializer(serializers.ModelSerializer):
@@ -22,6 +22,7 @@ class DirectoryEntrySerializer(serializers.ModelSerializer):
         read_only=True,
         many=True
     )
+    latest_scan = serializers.SerializerMethodField()
 
     class Meta:
         model = DirectoryEntry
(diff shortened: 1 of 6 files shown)


directory/api/serializers.py
```diff
@@ -38,4 +39,25 @@ class Meta:
...
+            'latest_scan',
         ]
+
+    def get_latest_scan(self, directory_entry):
+        latest = directory_entry.get_live_result()
+
+        if latest:
+            serializer_latest = ScanResultSerializer(instance=latest)
+            return serializer_latest.data
+        else:
```
SaptakS: Can skip the `else` since return statement.

SaptakS (commented):
In general, looks good to me. Left a small comment. The tests for API look good!

directory/api/serializers.py
chigby: Good point here, I've removed it.

SaptakS (approved):
Looks good to me

PR #767, opened 2020-11-02
Remove code related to legacy autocomplete

### LINKED ISSUE #766: Remove React on securedrop.org
It is looking to me like the only reason we have react in `package.json` is from when we were using the legacy, non-package autocomplete code. I think this can be removed, along with the react dependency, and that will fix our npm audit failures.

chigby:
This pull request removes all the code I could find that's connected to the legacy client-side autocomplete feature. The motivation for doing this is the presence of a [security vulnerability in a dependency of react](https://npmjs.com/advisories/1556).

The autocomplete code in our `client` directory has either been replaced by the `wagtailautocomplete` plugin on the admin side, or is related to the unused "directory management" feature.  This means I think it's safe to remove.  

It's possible we will want to remove the rest of the directory management feature at some point, but that I think requires some discussion.

Fixes #766


SaptakS (approved):
looks good to me

harrislapiroff (approved):

PR #768, opened 2020-11-04
Updates lxml and djangorestframework dependencies

SaptakS:
cryptography is a dependency of pyopenssl and sslyze. Even though pyopenssl's latest version has a >=2.8 for cryptography, but the sslyze version that we are using (because pshtt not compatible with more updated versions of sslyze) has a pinned dependency on ==2.5 of cryptography. Hence ignoring CVE for now.


chigby (approved):
Everything looks good here. I'm sad about the `pshtt` situation, but it looks like we're stuck for the time being. Thanks for getting these updates in!

PR #769, opened 2020-11-06
Remove settings related to HSTS, which is handled by nginx in production

maeve-fpf:
Since we have made HSTS be handled at the ingress-nginx level for consistency, these settings can be removed.


chigby (approved):
Not much else to say here but I think this looks good.

PR #771, opened 2020-11-12
HTTPS analytics

### LINKED ISSUE #754: Analytics loaded over HTTP on onion site
Pages on the SecureDrop site contain:

    <noscript><p></p></noscript>

Because the onion address is http:// rather than https://, this can load http://analytics.freedom.press. That causes a mixed-content warning in the "Safest" security level of Tor Browser since JavaScript is disabled. Can we fix that URL to use https:// ?

chigby:
This pull request replaces the use of `django-analytical` with our own code to insert the Piwik analytics javascript snippet with the appropriate variables customized. This change enables us to force the use of HTTPS when communicating with the analytics server.

Another change of note: now we can use a nonce (rather than a sha256 hash) on this code to allow it to pass the CSP.

Fixes #754

### DIFF
--- a/securedrop/static/js/piwik.js
+++ b/securedrop/static/js/piwik.js
@@ -1,0 +1,9 @@
+var idSite = 3;
+var piwikTrackingApiUrl = 'https://analytics.freedom.press/piwik.php';
+
+var _paq = window._paq = window._paq || [];
+
+_paq.push(['setTrackerUrl', piwikTrackingApiUrl]);
+_paq.push(['setSiteId', idSite]);
+_paq.push(['trackPageView']);
+_paq.push(['enableLinkTracking']);
--- a/securedrop/templates/super.html
+++ b/securedrop/templates/super.html
@@ -50,22 +50,24 @@
 	<body class="{% block body_class %}{% endblock %}">
 		{% block body %}
 		{% endblock %}
-		{% script type="application/javascript" async=False %}
-		<script type="text/javascript">
-			var _paq = _paq || [];
+		{% if django_settings.PIWIK_DOMAIN_PATH and django_settings.PIWIK_SITE_ID %}
+			{% script type="application/javascript" async=False %}
+			<script type="text/javascript">
+			 var _paq = _paq || [];
 
-			_paq.push(['trackPageView']);
-			_paq.push(['enableLinkTracking']);
-			(function() {
-				var u="https://{{ PIWIK_DOMAIN_PATH }}/";
-				_paq.push(['setTrackerUrl', u+'piwik.php']);
-				_paq.push(['setSiteId', {{ PIWIK_SITE_ID }}]);
-				var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
-				g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
-			})();
-		</script>
+			 _paq.push(['trackPageView']);
+			 _paq.push(['enableLinkTracking']);
+			 (function() {
+				 var u="https://{{ django_settings.PIWIK_DOMAIN_PATH }}/";
+				 _paq.push(['setTrackerUrl', u+'piwik.php']);
+				 _paq.push(['setSiteId', {{ django_settings.PIWIK_SITE_ID }}]);
+				 var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+				 g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
+			 })();
+			</script>
 		{% endscript %}
-		<noscript><p><img src="https://{{ PIWIK_DOMAIN_PATH }}/piwik.php?idsite={{ PIWIK_SITE_ID }}" style="border:0;" alt="" /></p></noscript>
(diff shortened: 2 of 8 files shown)


SaptakS (commented):
I wonder how frequently will the PIWIK DOMAIN PATH and PIWIK SITE ID change. If it's pretty much static and we don't care if the tracker loads in dev environment, I feel we can continue using sha-256 hash for CSP, given that we don't have many inline script, hence nonce might be overkill (since we are trying to allow only this one particular inline script). But if the script is kindof changing and dynamic, then nonce, should be the way to go.

I don't feel especially strong about either, so more or less fine with keeping it this way.

conorsch:
> I wonder how frequently will the PIWIK DOMAIN PATH and PIWIK SITE ID change.

The domain, almost never, at least looking historically. The ID, however, will change more frequently: first, if we add a new site, then we'll add a new ID. More to the point, in staging we try to set an incompatible ID so as not to pollute the prod analytics with page loads from the staging server. So that should at least be flexible for multiple environments. 

> If it's pretty much static and we don't care if the tracker loads in dev environment

I'm not certain that a load in the dev environment would actually submit a working event to the analytics server, but would certainly prefer to avoid it, as with staging. You raise an interesting point, @SaptakS : does this new logic support disabling analyitcs entirely for certain environments, such as dev? Looks to my eye like it does: https://github.com/freedomofpress/securedrop.org/blob/2d8ef7e05f1ed4e5006edc763102fcb527bc4ba0/securedrop/settings/base.py#L257-L268

chigby:
@conorsch I was actually just now taking a look at the implications of removing `django-analytical` on staging/dev environments. I think that now we have the capability to not render the analytics code at all if the two `PIWIK...` settings are set to `None`. To me this seems nicer than what we're doing now, and I can update this PR to include that feature if we agree it would be a good idea

conorsch:
So `None` would be the defaults for both values, and we only set them in prod? That sounds reasonable, although we would miss the opportunity to catch basic CSP problems with the JS load on staging. Honestly I don't think that's worthwhile to preserve, so the none defaults sound good to me.

chigby:
Okay, I've added the conditional rendering of the `<script>` tag.. although this is now making me consider if we can remove the need for using an inline script tag entirely by putting this code into its own file?

SaptakS:
> Okay, I've added the conditional rendering of the `<script>` tag.. although this is now making me consider if we can remove the need for using an inline script tag entirely by putting this code into its own file?

I agree. I feel we can definitely add it as a separate file avoding inline script altogether.

chigby:
I've made the change of moving the analytics code into its own file. @SaptakS let me know what you think. One consequence here is that we're not using environment variables anymore to configure the analytics site id or the URL, but I'm tending to think this is fine to code directly in the relevant JS file since we really don't expect it to change often (now that it will not be displayed in staging and dev environments).

maeve-fpf:
OK, so this deployment will need to remove the `PIWIK_*` variables and add `ANALYTICS_ENABLED`? Can you open an infra ticket?

chigby:
As it's configured right now, `ANALYTICS_ENABLED` is not set-up to get its value from an environment variable. The way it works is it's `True` if we're using the production settings file and `False` otherwise. Possibly this is inconsistent with the rest of our configuration, so I'm happy to change it. My thinking was that this flag didn't feel like the kind of thing that needed to be dynamically switched on/off at the environment level, but also I don't really have a strong basis or opinion on that. I guess the benefit would be that fewer variables is a simpler config? Let me know what you think.

I can open a ticket either way. Is https://github.com/freedomofpress/k8s-configs the repo I should be using for infra tickets?

maeve-fpf:
Sounds like a good idea to me! Yes, you can use that repo for tickets - in general, if something is removed from an app, we'd like to keep the configs there clean, so that will help track it.

securedrop/templates/super.html
```diff
@@ -1,4 +1,4 @@
-{% load static wagtailmetadata_tags wagtailsettings_tags common_tags piwik %}
+{% load static wagtailmetadata_tags wagtailsettings_tags common_tags csp %}
```
SaptakS: Correct me if I am wrong but I don't think we need to load `csp` any longer? Might also no longer need to include csp in INCLUDED_APPS if I am not wrong.

SaptakS (commented):
Mentioned a small change. Everything else looks good to be merged.

securedrop/templates/super.html
chigby: I think you are right, here. I was using `csp` for the `script` tag, which is not there anymore. I'll remove it.

chigby:
Thanks for the final review. I have updated the code to remove the extra loading of `csp` as an installed app and as a template tag.

securedrop/static/js/piwik.js
```diff
@@ -0,0 +1,9 @@
+var idSite = 3;
+var piwikTrackingApiUrl = 'https://analytics.freedom.press/piwik.php';
```
harrislapiroff, line 2: I'm a little on the fence about hard coding these values. Apart from them philosophically feeling like _configuration_ there was at least one occasion when I needed to test analytics recording on a local Matomo instance, which would be a little trickier with this set up.

That having been said, this is fine for now and we can revisit later.

harrislapiroff (approved):
This all looks reasonable to me, let's pop it onto the sandbox!

PR #773, opened 2020-11-19
Should fix analytics JS code ordering

### LINKED ISSUE #772: New Analytics code causing errors
I'm getting this error on staging with the new analytics code

> _paq.push() was used but Matomo tracker was not initialized before the matomo.js file was loaded. Make sure to configure the tracker via _paq.push before loading matomo.js. Alternatively, you can create a tracker via Matomo.addTracker() manually and then use _paq.push but it may not fully work as tracker methods may not be executed in the correct order. 

I think we might need to ensure the proper order of execution on our javascript either by moving the analytics script back to the HTML or maybe by not loading the JS asynchronously? Will investigate the docs.

SaptakS:
Should fix #772 

Based on the error and the docs, I think the reordering should do the trick. The `async` should not affect because the previous script isn't `async` so that script will need to be loaded, and only `https://analytics.freedom.press/piwik.js` will load asynchronously which should be fine.

### DIFF
--- a/securedrop/templates/super.html
+++ b/securedrop/templates/super.html
@@ -52,8 +52,8 @@
 		{% endblock %}
 
 		{% if django_settings.ANALYTICS_ENABLED %}
-			<script src="https://analytics.freedom.press/piwik.js" async defer></script>
 			<script type="text/javascript" src="{% static 'js/piwik.js' %}"></script>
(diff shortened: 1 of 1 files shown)


chigby (approved):
Good catch here! Let's get this onto staging for testing.

PR #774, opened 2020-11-19
Tor browser fixes

harrislapiroff:


SaptakS (approved):
The regex and sass changes look good to me.

PR #779, opened 2020-12-07
Model admin for scan results

### LINKED ISSUE #777: Directory editor takes seconds to load, hits error 400 on submit
## Steps to reproduce
1. Edit Al Jazeera's directory entry https://securedrop.org/admin/pages/125/edit/
2. Input the correct landing page: https://webapps.aljazeera.net/aje/tips/ into the "landing page" field
3. Click "save draft"

## Expected behavior
- Directory editor loads quickly
- After "Save draft", the draft is updated

## Actual behavior

- Editor page takes many seconds to load
- Saving draft results in HTTP 400 error without further details. 

## Additional information

Log example:

```
{"request":"<WSGIRequest: POST '/admin/pages/125/edit/'>","status_code":400,"module":"exception","name":"django.security.TooManyFieldsSent","levelname":"ERROR","message":"The number of GET/POST parameters exceeded settings.DATA_UPLOAD_MAX_NUMBER_FIELDS.","asctime":"2020-12-03 22:03:45,459"}
```

This may be due to the way all attempted landing page scans are currently associated with the directory entry.

chigby:
This pull request replaces the inline view of scan results that appears on the directory entry admin page with a separate Model Admin area. The reason for this is the inline scan result list becomes slow if there are a large number of results, and also can potentially break the edit functionality.

Instead of that, I've added a drop down button on the page listing for directory entries:

image

Fixes #777

### DIFF
--- a/directory/wagtail_hooks.py
+++ b/directory/wagtail_hooks.py
@@ -68,7 +68,7 @@ def add_scan_results_buttons(page, page_perms, is_parent=False):
 @hooks.register('add_scan_results_button')
 def add_bundle_stats_button(page, page_perms, is_parent=False):
     index_url = scanresult_modeladmin.url_helper.get_action_url('index')
-    results_url = index_url + '?securedrop__page_ptr_id__exact=' + str(page.pk)
+    results_url = f'{index_url}?securedrop__page_ptr_id__exact={page.pk}'
     latest = page.get_live_result()
     return [
         Button(
--- a/directory/wagtail_hooks.py
+++ b/directory/wagtail_hooks.py
@@ -31,6 +31,7 @@ class ReadOnlyPermissionHelper(PermissionHelper):
 
 
 class ScanResultAdmin(ModelAdmin):
+    """ModelAdmin for viewing/searching ScanResults."""
     model = ScanResult
     menu_icon = 'folder-open-inverse'
     menu_order = 500
@@ -40,7 +41,7 @@ class ScanResultAdmin(ModelAdmin):
         'grade',
         'live',
     )
-    search_fields = ('landing_page_url',)
+    search_fields = ('landing_page_url', 'securedrop__title')
     permission_helper_class = ReadOnlyPermissionHelper
     inspect_view_enabled = True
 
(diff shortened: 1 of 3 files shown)


directory/wagtail_hooks.py
```diff
@@ -0,0 +1,87 @@
...
+            page_perms=page_perms,
+            is_parent=is_parent,
+            priority=35
+        )
+
+
+@hooks.register('add_scan_results_button')
+def add_bundle_stats_button(page, page_perms, is_parent=False):
+    index_url = scanresult_modeladmin.url_helper.get_action_url('index')
+    results_url = index_url + '?securedrop__page_ptr_id__exact=' + str(page.pk)
```
SaptakS: I prefer `.format()`. In fact since we have updated to python3.7, we should even be able to use `f'{index_url}?securedrop__page_ptr_id__exact={page.pk}'`. Just makes better readability according to me.

Also, maybe we can add docstring to this hook and the ScanResultAdmin class as well?

directory/wagtail_hooks.py
```diff
@@ -0,0 +1,87 @@
...
+    model = ScanResult
+    menu_icon = 'folder-open-inverse'
+    menu_order = 500
+    list_display = ('securedrop', 'result_last_seen', 'live', 'grade')
+    list_filter = (
+        'result_last_seen',
+        'grade',
+        'live',
+    )
+    search_fields = ('landing_page_url',)
```
SaptakS: Curious if we should allow also the `securedrop` field for searching?

SaptakS (commented):
Looks good to me. Left some comment on code basis. Can be merged after that.

Just to note here, the scan results are no longer visible in the edit page. One will have to click the scan result dropdown button from the listing page. Also, they can directly go to the Scan Results and search by entering the landing page url. I think this works and is a substantial improvement but might be something to note and remember. cc @eloquence

chigby:
Thanks for the feedback here.

1. Definitely prefer the `f''` string approach here. Need to get into the habit of using that!
2. Added some additional docstrings.
3. Allow searching by directory entry title
4. Added a helpful link section at the bottom of the DirectoryEntry editor (where the big list of all the results was previously), so now you can preview and quickly jump to the results no matter where you are in the admin:

image

SaptakS (approved):
looks good to me.

PR #786, opened 2020-12-15
Make some settings consistent with other apps

### LINKED ISSUE #778: Make SDO settings consistent with other apps
There are a few quirks in each app's settings that we'd like to make a little more unified. Things we can do on SDO:

- Adjust how CSP variables like `DJANGO_CSP_*_HOSTS` are used: variable should contain the full entry (with `https://`), and if possible, consolidate the vars into one. Other apps should follow the example here.
- Instead of `DJANGO_LOG_CONSOLE` and `DJANGO_LOG_PATH`, just log to console if a log path is not set. Don't require `DJANGO_LOG_JSON`, default it appropriately.
- Treat env vars that should be "booleans" the same everywhere.

maeve-fpf:
Works towards #778 

This is similar to https://github.com/freedomofpress/securethenews/pull/299, but for SDO. Like that, the diff is pretty large due to some dependency housekeeping.

We never pulled in a new `django-logging` to include https://github.com/freedomofpress/django-logging/pull/4/files here like we did in STN, so upgrade that as well.


chigby (approved):
Changes look good to me, very similar to the situation over on STN and PFT and they seem to work here as described.

PR #789, opened 2020-12-16
Adds onion names for directory entries

### LINKED ISSUE #782: Create form option for onion name in SecureDrop directory entry
Now that we have [onion names](https://securedrop.org/news/introducing-onion-names-securedrop/), it would be good to show them in an instance's directory entry. We should show both the onion name and the 56-char v3 url, to reinforce that both are valid for an instance.

Not every instance will have an onion name, so it should be an optional field.

Happy to provide more details as needed!

SaptakS:
Fixes #782 

Onion names field is added in the directory entry's admin page.
It is shown in both the directory entry page and also in the API results.

Just one little thing to add. I have added a regex check for the format of the onion name to be as `.securedrop.tor.onion`.


directory/models/entry.py
```diff
@@ -106,6 +106,14 @@ class DirectoryEntry(MetadataPageMixin, Page):
         validators=[RegexValidator(regex=r'\.onion$', message="Enter a valid .onion address.")]
     )
 
+    onion_name = models.CharField(
+        'SecureDrop onion names',
```
chigby: Kind of a nitpick here, but this field feels to me like it should be "name" instead of "names". Unless there's a way to add multiples?

directory/models/entry.py
```diff
@@ -106,6 +106,14 @@ class DirectoryEntry(MetadataPageMixin, Page):
...
+        max_length=255,
+        null=True,
+        blank=True,
+        validators=[RegexValidator(regex=r'\.securedrop\.tor\.onion$', message="Enter a valid onion name.")]
```
chigby: This is possibly just my lack of familiarity with the subject matter showing, but ".securedrop.tor.onion" is not obvious to me, so I'm kind of inclined here to want to offer the user some kind of explanation here for what constitutes a valid onion name, either in the error message on the validator or as help text on the field itself. 

It's definitely possible that the `onion_address` field above this one could be improved in the regard, too. :sweat_smile:

directory/templates/directory/_instance_instructions.html
```diff
@@ -8,12 +8,19 @@ <h1 class="instance-instructions__title">{% trans "Planning to submit informatio
 			we recommend that you compare the .onion address provided on the
 			organization’s landing page to this one
 			<span class="instance-instructions__onion-address">{{ onion }}</span>
+		{% endblocktrans %}
+		{% if page.onion_name %}
+			{% blocktrans with onion_name=page.onion_name %}
+				<br />[or <span class="instance-instructions__onion-address">{{ onion_name }}</span>]
```
chigby: In reading the ticket, it looks like this entire block of text was reworked to be shorter. It'd be nice to get those changes in here as well.

directory/tests/factories/entry.py
```diff
@@ -21,6 +21,11 @@ def random_onion_address():
     ) + '.onion'
 
 
+def random_onion_name():
+    name = factory.Faker('sentence', nb_words=1)
+    return f'https://{name}.securedrop.tor.onion'
```
chigby, line 28: You added the onion name attribute to the factories! I think that's great. The factories are not always kept up-to-date with the models. 

I do have to offer a correction here, since as you have written it I don't think it works as intended (or at least it didn't for me locally). 

The problem is `factory.Faker(...)` does not return fake data directly, it's what you declare on a factory class to tell that factory to use fake data for that attribute. So your `name` variable here does not contain a name.

There are basically two ways around this. One is to use the faker library directly, _without_ it being part of factory:

```python
from faker import Faker

fake = Faker()
name = fake.word()
```
Then you'll have an actual fake word in your random function.

The other way is more declarative (and potentially more verbose) but I've used it a fair bit in the tracker repo. Essentially what you do is define the fake data as an attribute on the factory that's not part of the model, just intended to be used by other attributes. Here's a [fairly concise example](https://github.com/freedomofpress/pressfreedomtracker.us/blob/c6731b8c04b89dd30b7f58f0298dca13a7be0452/incident/tests/factories.py#L408-L420). I like this method because keeps everything to declarations within the factory, but it might be more machinery than we need here, especially since we already have most of the code for the other method.

directory/tests/test_models_entries.py
```diff
@@ -33,6 +34,12 @@ def test_securedrop_cannot_save_invalid_onion_address(self):
                 onion_address='https://notreal.com',
             )
 
+    def test_securedrop_cannot_save_invalid_onion_name(self):
+        with self.assertRaises(ValidationError):
+            DirectoryEntryFactory(
+                onion_name='https://notreal.com',
+            )
```
chigby, line 41: :+1:

chigby (changes requested):
Left some comments below. I don't think these changes are too much, but it is looking like we are not quite ready to merge this yet. Let me know what you think about this, or if I have something wrong there.

directory/templates/directory/_instance_instructions.html
SaptakS: I did add the other text changes mentioned in the ticket as well. 🤔  [This](https://github.com/freedomofpress/securedrop.org/pull/789/files/44cbf07146f16b3f91e18006736fc2c974ded8de#diff-bf1b9c020538c9f24962d6f8d3c21ec616d179aa469a3158fbda3f72e4f63817R21) and [this](https://github.com/freedomofpress/securedrop.org/pull/789/files/44cbf07146f16b3f91e18006736fc2c974ded8de#diff-bf1b9c020538c9f24962d6f8d3c21ec616d179aa469a3158fbda3f72e4f63817R23)

chigby (approved):

PR #793, opened 2020-12-21
Adds word break to onion address

### LINKED ISSUE #792: Mobile UI breaks for directory listing with v3 onion address
Because of the longer v3 onion address, the directory listing in mobile screens break. They have a horizontal scroll.

For example: https://securedrop.org/directory/new-york-times/

SaptakS:
Fixes #792 
Adds word break to onion address so that directory listing with v3 onion address don't break the UI in smaller screens.


chigby (approved):
Works great, fixes the problem perfectly.

Issue #784, opened 2020-12-14
Mastodon link verification fails, potentially due to securedrop.org page size


eloquence:
See https://github.com/tootsuite/mastodon/issues/15316#issuecomment-744080451

eloquence:
Upon quick inspection, the inline SVG hexagon background appears to be a large chunk of this, though other inline SVGs or resources may as well. As a matter of practice, perhaps we should define a cutoff size we use to decide whether it's appropriate to load a resource inline vs. as a static asset?

eloquence:
It was indeed due to the large page size and we're now verified on Mastodon.

PR #794, opened 2020-12-29
Uses separate hexagon files for inline svg rendering

SaptakS:
Since we are rendering the svg inline, I think we can use separate svg for different sections based on the requirement. A lot of the areas hae the hexagons scaled up, hence not needing all the hexagons. Hence, I have broken them down into different svg files. The browser effectively doesn't need to load different svg files (since inline rendered), so performance shouldn't be affected. I did a local performance test using chrome lighthouse and was just slight improvement (mostly because of the reduced html size)

The current html generated after this changes is 666.33kB compared to the previous 1.72MB

Refs: #784


SaptakS:
PS: There are similar svg used in other pages also because of which each page is almost greater than 1MB in size (since the svg is around 400kB in size)

chigby (approved):
I checked this locally and the home page size is reduced as described, and also it looks the same to my eyes, so, well done!

Do we want to do the same to the other pages in this PR? I'm fine with merging this now and saving those for later, but if you want to, I'm fine with waiting for the others also.

PR #795, opened 2021-01-04
Fix button rendering bug in directory entry admin listing

chigby:
This fixes a bug on the directory entry admin listings where if the latest scan result for a directory entry in the listing does not exist, then the button creation hook would raise an exception and the admin page would not render.  I've fixed it by only adding that button if the latest result actually is present.

I was running into this while doing some testing locally, but I don't see why it wouldn't happen in production eventually, either.

### DIFF
--- a/directory/wagtail_hooks.py
+++ b/directory/wagtail_hooks.py
@@ -76,19 +76,23 @@ def add_bundle_stats_button(page, page_perms, is_parent=False):
     """
     index_url = scanresult_modeladmin.url_helper.get_action_url('index')
     results_url = f'{index_url}?securedrop__page_ptr_id__exact={page.pk}'
-    latest = page.get_live_result()
-    return [
+    buttons = [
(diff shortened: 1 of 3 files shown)


directory/wagtail_hooks.py
```diff
@@ -76,19 +76,23 @@ def add_bundle_stats_button(page, page_perms, is_parent=False):
...
             priority=10,
-        ),
-        Button(
-            'Latest Result',
-            scanresult_modeladmin.url_helper.get_action_url('inspect', latest.pk),
-            priority=20,
         )
     ]
+    latest = page.get_live_result()
+    if latest and latest.pk:
```
SaptakS, line 87: Makes sense. I remember I added a similar check for a new directory page where latest wouldn't be there. Might have missed this one at that time.

SaptakS (approved):

PR #798, opened 2021-01-12
Changes inline styles to attributes in inline-svgs

SaptakS:
_hexagon.svg and _logo_color.svg both used inline style to add `fill` and `fill-rule` both of which can also be added as attributes to the elements itself in svg. Since chrome sometimes have some issues with sha based CPs (https://bugs.chromium.org/p/chromium/issues/detail?id=546106), the securedrop logo was breaking in mobile chrome browsers.

By moving from inline-styles to  attributes, there is no need to depend on CSS CSPs, to load the svg properly. Also, in the process of removing other inline styles, I removed `style="border:0;"` in piwik noscript link because it seemingly did nothing as such.

PS: there are other svgs with inline styles but I verified that all of them are being loaded in the `` tag instead of rendering inline svg, so it doesn't affect.


SaptakS:
Note for future self, and others in the web team:  To add hashes to allow inline styles, chromium needs that the hashes be added to `style-src-attr` along with `'unsafe-hashes`

So, in our code, it needs to be something like:
```python
CSP_STYLE_SRC_ATTR = (
    "'unsafe-hashes'",
    "'sha256-<hash>'",
)
```

in addition to the `CSP_STYLE_SRC`

Refs: https://chromium-review.googlesource.com/c/chromium/src/+/1527526

Might be worth adding it in other websites, in case we are facing CSP issues.
cc: @chigby @harrislapiroff

chigby (approved):
This is a really great solution to one of our CSP problems. To my eye the SVG renders identically and without any of the need to maintain inline style hashes.

PR #808, opened 2021-02-16
Update wagtail

### LINKED ISSUE #805: Upgrade to latest Wagtail LTS

SaptakS:
Resolves #805

Following are the changes in the code that were needed:
- Updating cloudflare API key name
- Enable legacy rich text behaviour
- migration for clean_name
- change from page_type to target_model in the updated wagtail-autocomplete

The new features are documented [here](https://github.com/freedomofpress/pressfreedomtracker.us/pull/1002#issue-553720664) for the PFT upgrade by @chigby which is similar for this PR as well for the most part


chigby (changes requested):
I think all of this looks good, and it seems like the same strategies from tracker applied here as well. Perhaps we ought to think about updating the rich text CSS to not require the legacy mode across our sites.

One thing I did notice is that apparently between the time when this PR was made and today, wagtail 2.11.5 was released (this PR updates to 2.11.3) so maybe we should take the opportunity to get the latest in there?

harrislapiroff:
Agree, let's get this up to the latest. Thanks, both!

SaptakS:
Updated to 2.11.5. Should I try to implement new designs and deprecate the legacy mode as part of this PR itself? I am guessing that is something we will need to implement more or less in all the web repos.

chigby:
IMO we should not try to update the styles in this PR. Sorry if the earlier comment made it seem like I thought we should do them now.

chigby (approved):

PR #812, opened 2021-02-23
Fix rich text rendering visual errors across site

harrislapiroff:
We rely on wagtail.contrib.legacy.richtext to accomodate our styles
which rely on legacy rich-text rendering, but the contrib needs to
be placed before wagtail.core for legacy rendering to take
precedence. Now it is.


SaptakS (approved):
LGTM

PR #834, opened 2021-04-07
Remove discourse search

### LINKED ISSUE #800: Remove forum search integration
The forum is largely inactive except for translation announcements and the occasional support question (which typically gets redirected to https://support.freedom.press/ quickly). The securedrop.org search currently covers the forum, which requires maintenance (e.g., #740). 

Overall, the securedrop.org search page receives <50 views in a given month, and it's highly doubtful this forum integration provides meaningful utility to SD users; more likely, it actually makes documentation and site content easy to miss among outdated forum posts.

At this point, the forum search feature can be removed (the search placeholder can just say "Search website and documentation").

chigby:
Fixes #800 

This pull request removes discourse/forum related search code from the repo entirely. It also adds a data migration to remove any search documents with the type of `F` from the database. In theory this could be done manually after this branch is merged, but I thought this might be easier. Let me know if there is a different preferred way of removing those documents.


SaptakS (changes requested):
I think there are few other places that need to be updated, like the templates in `search/templates/search` and also there is a svg file for forum icon in search results which can be removed I think.

SaptakS (approved):
Looks good to me

PR #837, opened 2021-04-12
Replaces node-sass with dart-sass

SaptakS:
Refs https://github.com/freedomofpress/fpf-www-projects/issues/173


chigby (approved):
Everything checks out on my end after installing the new sass and related packages. The code builds as it did before, and it's nice we even seem to have caught some errors that were not being seen (or were silently being fixed, who knows) by the old version.

PR #838, opened 2021-04-12
Remove "www." prefix from domains before inspecting with pshtt

### LINKED ISSUE #739: Scan comes up with invalid domains
I have noticed that the scan is logging errors about `www.www.washingtonpost.com`, which does exist in DNS (see https://subdomainfinder.c99.nl/scans/2020-09-26/washingtonpost.com) but is not configured to serve a valid TLS certificate. I don't see a reference to anything but `www.washingtonpost.com` in the relevant directory entry, and the front page does not appear to include any assets from it.

We could:

- add an ignore for this specific case
- try to contact the org and see if they really need this (and other, similar misconfigured subdomains) to still be running
- ignore any hosts that aren't sourced from the page scanned

FPF folks, I can send/refer to logs, but it seems like a good idea to not post them here.

chigby:
Fixes #739

This pull request adds some code to remove, if present, the "www." prefix from any domain name given to pshtt. Pshtt has its own logic to test the www and non-www versions of the domains its given, and it expects to be given the base domain, otherwise it will add "www." on is own, duplicating it if it's already there.

### DIFF
--- a/scanner/scanner.py
+++ b/scanner/scanner.py
@@ -47,7 +47,11 @@ def perform_scan(url: str, permitted_domains: List[str]) -> ScanResult:
     asset_results = parse_assets(assets, [tldextract.extract(page.url).registered_domain] + permitted_domains)
     scan_data.update(asset_results)
 
-    pshtt_results = list(inspect_domains([url_to_domain(page.url)], {'timeout': 10}))
+    pshtt_domain = url_to_domain(page.url)
+    if pshtt_domain.startswith('www.'):
+        pshtt_domain = pshtt_domain[4:]
+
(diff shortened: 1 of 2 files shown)


SaptakS (commented):
Looks good to me. I am just wondering if it fails similarly for other subdomains also? Don't know if we have such cases. But for example if there was sub.example.com, it will try to check www.sub.example.com? At least looks like that from the README of pshtt

chigby:
That was my assessment as well: if given sub.example.com, it will try www.sub.example.com. I am not sure if this is undesired, though.

SaptakS (approved):
I am going to merge this and we can address if any false positives come up due to subdomains logic.

PR #845, opened 2021-05-04
Switch pshtt to forked version

### LINKED ISSUE #844: Switch to fork of `pshtt` to update `sslyze`, `cryptography`
`pshtt` is stuck on an older version of the `sslyze` dependency, which prevents us from updating other dependencies as well, including `cryptography`. We've submitted an upstream PR to resolve (https://github.com/cisagov/pshtt/pull/214), but it's moving very slowly, so we set an internal deadline to switch to a fork temporarily. That deadline was May 1, which has now passed.

After creating a `freedomofpress` fork, let's update the associated dependencies in this repo.

chigby:
This pull request switches `pshtt` to the version of it we have forked. Also upgrades related packages where possible, namely `cryptography`, which means we can also un-ignore all the safety checks we have been previously ignoring.

Fixes #844

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,12 +4,6 @@ anyascii==0.1.7 \
     # via
     #   -r requirements.txt
     #   wagtail
-asn1crypto==1.4.0 \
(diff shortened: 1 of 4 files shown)


SaptakS (approved):
LGTM

PR #848, opened 2021-05-11
Update Matomo/Piwik tracking code to match latest recommendation

### LINKED ISSUE #809: Update Analytics tracking code
Looks like the current Matomo tracking code looks like this. It's not so different from what we have now, but it might be worth updating just in case:

```html
<!-- Matomo -->
<script type="text/javascript">
  var _paq = window._paq = window._paq || [];
  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
  _paq.push(['trackPageView']);
  _paq.push(['enableLinkTracking']);
  (function() {
    var u="https://analytics.freedom.press/";
    _paq.push(['setTrackerUrl', u+'matomo.php']);
    _paq.push(['setSiteId', '3']);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.type='text/javascript'; g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
  })();
</script>
<!-- End Matomo Code -->
```

N.b., adding this code will also involve removing the script tag in the template loading `https://analytics.freedom.press/piwik.js`

chigby:
This pull request updates the Matomo tracking code to match the latest recommended use of their tracking code. Notably, the loading of the `https://analytics.freedom.press/piwik.js` resource is moved into the JS file itself, instead of being loaded via a script tag.

Fixes #809

### DIFF
--- a/securedrop/static/js/piwik.js
+++ b/securedrop/static/js/piwik.js
@@ -1,9 +1,11 @@
-var idSite = 3;
-var piwikTrackingApiUrl = 'https://analytics.freedom.press/piwik.php';
-
 var _paq = window._paq = window._paq || [];
-
-_paq.push(['setTrackerUrl', piwikTrackingApiUrl]);
-_paq.push(['setSiteId', idSite]);
+/* tracker methods like "setCustomDimension" should be called before "trackPageView" */
 _paq.push(['trackPageView']);
 _paq.push(['enableLinkTracking']);
+(function() {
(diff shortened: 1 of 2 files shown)


SaptakS (approved):
Just want to make a note that it is best to keep an eye on the tracking maybe on staging to see that merging of this code didn't lead to anything breaking.

PR #849, opened 2021-05-12
Show directory search query in the search form

### LINKED ISSUE #814: Directory search form not preserving search query
Query data is being properly put in GET params and searches are being executed properly, but the search term should be seeded back into the form as an initial value when the page loads. Currently it isn't—see the search field show up blank when I execute the search below:

https://user-images.githubusercontent.com/296631/109051084-24dd5380-76a8-11eb-8405-fda388e4b0d2.mov

chigby:
Fixes #814

This pull request grabs the search query from the GET params and puts it into the search form.


SaptakS (approved):
Works perfectly for me.

PR #854, opened 2021-05-26
Change django-logging encoding from 'ascii' to 'utf-8'

chigby:
The default is not acceptable for logging the "content" response, which may contain non-ascii characters.  Even though we are ultimately discarding the content, `django-logging` still encodes it first.

Refs https://github.com/freedomofpress/fpf-www-projects/issues/180


SaptakS (approved):
looks good to me

Issue #803, opened 2021-02-04
Make Landing Page scan a pre-requisite for submitting directory listing? 


rocodes:
Hi all,

As we transition to v3 onion services, the idea of onion names will be more and more appealing, and we're hopefully going to get more requests for directory listings and onion names as a result (already happening in modest numbers).

In order to get instances listed more quickly and also reduce support burden (corresponding explaining landing page guidelines etc), it could be a good idea to allow (or require?) instances to try a scan before submitting a directory listing.  There seems to be a lack of clarity about implementing the guidelines, or a lack of transparency that the 'guidelines' are actually requirements before being listed in the directory, and a fair amount of time is spent corresponding with instances and their web teams to get them into compliance before approving their listing.

This might require a few UI tweaks to indicate which criteria are required (e.g. CSP, HSTS, Referrer Policy) and which ones are preferred but not required.

harrislapiroff:
Ooooh, this is an interesting idea. We'd have to think at least a little carefully about how to implement it since a scan can be quite resource consuming and we definitely don't want someone just hammering our server requesting scans, but I like it as a concept! Maybe with a rate limit and possibly a captcha, this could work?

@eloquence, what do you think?

eloquence:
I like the idea as well. I would suggest implementing in stages:
- only on staging site
- as a standalone page linked to from the submission form
- (potentially) integrated with the submission form

@rocodes If you'd like to help move this forward, I think prioritizing the information we'd want to surface would be the next step to make this ticket more actionable.

rocodes:
Yes, I was also wondering about the rate-limiting or other anti-abuse strategies.

Here is the matrix we currently use to evaluate instances. In addition to this, we also scan sites (e.g. using SSLLabs) to ensure up-to-date cipher suites. The items with "E" (enforce) are the ones that we require, so I would say we should absolutely surface those (if they can be handled by the scanner and aren't a manual check), and then the rest of them are optional/improvements.

(So for clarity: things to surface for sure would be HSTS / HSTS details, CSP, Referrer policy; things that would be nice to surface would be checking for 3rd-party resources and scripts, scanning for hyperlinks to securedrop.org or to onion urls).      

| Security Feature | Expected Result | Actual Result | Pass/Fail | Severity |
| ---------------------------------------- | ---------------------------------------- | ---------------------------------------- | --------- | --------- |
| HTTPS on landing page | yes |  |  | E |
| Loads in Tor Browser | yes | | | E |
[...]

PR #858, opened 2021-06-09
Manual scans of SecureDrop landing pages

chigby:
Refs #803 

This pull request adds a "Perform a Scan" button to the "Scan Results" section of the admin, as shown in the top right corner:

image

From there, you can supply the needed information and the scan will be done based on what you tell it:

image

The scan might take a little white to complete, but when it is, you will be redirected to the "Inspect" view for that result. For example:

image

This feature is available to any user who has "Create" permissions for Scan Result objects, which can be configured on a per-group basis.

I've tried to do things mostly in line with how Wagtail's Model Admins work, with the manual scan serving the role of the "create" action on `ScanResult` models, which I think actually works out fairly nicely. There are some more technical details I've written a little bit about in the commit messages.

### DIFF
--- a/directory/forms.py
+++ b/directory/forms.py
@@ -6,3 +6,15 @@ class ScannerForm(forms.Form):
     required_css_class = 'basic-form__required'
 
     url = forms.URLField()
+
+
+class ManualScanForm(forms.Form):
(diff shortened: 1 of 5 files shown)


SaptakS (approved):
Looks good to me.

PR #859, opened 2021-06-09
Log to the console by default

chigby:
This pull request changes the log configuration settings so that logs are directed to the console by default. Logs will only be sent to a file if `DJANGO_LOG_DIR` is configured. I believe this is how our system was intended to work and it was changed inadvertently.


harrislapiroff (approved):
Looks good in code, let's get it up

PR #878, opened 2021-07-19
Adds https_preferred checkbox for directory entries

### LINKED ISSUE #832: Consider tracking preferred protocol for onion addresses
Currently we do not track whether a SecureDrop uses HTTP or HTTPS. Given that onion services have their own encryption layer, this is generally not a security concern, but see https://community.torproject.org/onion-services/advanced/https/ for the pros and cons of running onion services over HTTPS.

We could potentially set the protocol in cases where a SecureDrop is served over HTTPS. This would also allow the [automated script that populates the onion name ruleset](https://github.com/freedomofpress/securedrop-https-everywhere-ruleset/blob/main/sddir.py) to use this protocol, instead of [always using HTTP](https://github.com/freedomofpress/securedrop-https-everywhere-ruleset/blob/1b5122653911135446a2df598310bee37b20669e/sddir.py#L10).

This appears to be mostly a content issue as the current validator only checks for the `.onion` suffix and will happily accept any protocol:
https://github.com/freedomofpress/securedrop.org/blob/6e6274dfc0fcb25106464828489b68dabf3fd746/directory/models/entry.py#L103-L107

The API, too, will return results with or without protocol, and the directory entries will show the protocol if it is included.

Advantages of tracking the protocol:
- We avoid an HTTP->HTTPS redirect when using an onion name or copy/pasting the address, which could mitigate against downgrade attacks by a very sophisticated attacker.

Disadvantages:
- If a news org switches to HTTPS and then stops using it, onion names will break (bug or feature?)

SaptakS:
If https_preferred is checked, no matter if the onion_address
has http, https or nothing as prefix, it will always add https://
as prefix. If not checked, it will just show whatever was entered
in the onion_address

Resolves #832


SaptakS:
I was wondering if we want to enforce the "HTTPS Preferred" on onion_names as well, or just the onion_address. cc @eloquence

SaptakS:
This also updates `pillow`

chigby:
The code itself looks good, though I can't speak to the design question.

harrislapiroff (approved):

PR #885, opened 2021-08-16
Validate html

### LINKED ISSUE #876: Address HTML validation issues
We need to address HTML validation issues in the code:

- [ ] https://validator.w3.org/nu/?doc=https%3A%2F%2Fsecuredrop.org%2F

- [ ] https://validator.w3.org/nu/?doc=https%3A%2F%2Fsecuredrop.org%2Foverview%2F

- [ ] https://validator.w3.org/nu/?doc=https%3A%2F%2Fsecuredrop.org%2Fdirectory%2F

- [ ] https://validator.w3.org/nu/?doc=https%3A%2F%2Fsecuredrop.org%2Fnews%2F

- [ ] https://validator.w3.org/nu/?doc=https%3A%2F%2Fsecuredrop.org%2Fcontribute%2F

- [ ] https://validator.w3.org/nu/?doc=https%3A%2F%2Fsecuredrop.org%2Ffaq%2F﻿

SaptakS:
Fixes #876 

There are certain HTML validations that wasn't fixed since they are being added by wagtail and wagtail-metadata:
- We want our bundle JS to be at the end, and because of that, it has to be kept as a stray tag (i.e. outside the <body> tag). If we move it inside the body tag, then wagtail admin adds a bunch of other things after that when logged in. Now this may or may not have any affect. Needs further investigation
- The itemprop attribute was specified, but the element is not a property of any item for meta tags. This is done in wagtail-metadata, and is a valid microdata formatting for SEO meta tags even though it seems to be showing up as an invalid HTML. I am investigating further to see if anyone else has done anything about.
- [This commit](https://github.com/freedomofpress/securedrop.org/pull/885/commits/d200128d0d75bc273ff3c92c07ea5088b03fd984) I didn't like doing but had to for now since we seem to be wanting richtext inside a heading tag. A [heading tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) can have only [phrasing contents](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content) as child, but not tags like `<div>`, `<p>`, etc. I would like to find a UX solution to hopefully go back to using h2.


SaptakS:
I am still kind of researching if we can add some html validation in our CI. Was trying https://github.com/svenkreiss/html5validator but it won't work on localhost:8000. The java tool which the python tool is a wrapper of, seems to work on HTML stdin output, so maybe I can pipe the output to that and see if that works in CI.

chigby:
One idea for validating in CI: write a test for it. I could imagine something similar to this: https://github.com/freedomofpress/freedom.press/blob/7027981aaf18d551fd0fba5f71ddf406d44a98ee/core/management/commands/tests/test_createdevdata.py#L12-L27 where instead of making a request we run the HTML of every page made by createdevdata through a validator of some kind.

SaptakS:
Yeah! That would be great. Though the tool itself is supposed to be used as a cli, but maybe I can import the class and run it on the response.content or something around that idea.

cfm:
Mentioned this in Slack but wanted to flag here too in case it's useful or worth discussing cross-team:  I've proposed a similar implementation for SecureDrop in freedomofpress/securedrop#6072.  :-)

SaptakS:
@cfm I did check that PR. We don't have integration tests, so saving will not be a good idea (we will need to curl or save reponse.content to do that anyways I think). So better to pipe output. But the piped HTML isn't working with html5validator based on my testing. But I think I like the path @chigby  mentioned in the PR. Might be worth checking if we can just import the Validator class and do run_vnu maybe. Else, there is still the option of running directly the java tool. Last resort I think would be just to curl and save the html files in a temporary location.

As discussed with @harrislapiroff though, I think it's probably a good idea to tackle this in a separate PR, and get this reviewed first. cc @chigby

common/templates/common/_footer.html
```diff
@@ -41,7 +41,15 @@ <h2 id="aria-support-nav" class="support__title">{{ footer.support_title }}</h2>
...
-					<h2 id="aria-footer-nav" class='footer-main__title'>{{ footer.title|richtext }}</h2>
+
+					<!--
+						It hurts me to make this change a little but wagtail legacy richtext always adds a div.
+						If we remove legacy richtext at some point also, there will be a p tag. What we technically
+						want is if we want only a link always in footer and no other usecase for richtext, then
+						just provide an option to add the link, instead of providing an entire richtext. That way,
+						we can move back to using h2. Right now this is the most valid solution, though not semantic :(
+					-->
+					<div role="heading" aria-level="1" id="aria-footer-nav" class='footer-main__title'>{{ footer.title|richtext }}</div>
```
chigby, line 52: Yeah, I think I agree with what you have done here given the circumstances while still preserving the graphic design. It is definitely problematic to put a richtext block inside a `h2` tag, and we should probably make an issue or something for changing this. To be honest, this footer text feels like something that could even be hard-coded. Or split into two fields like "parent organization name" and "parent organization link" if we feel we really might need to change those things in the admin. I guess another option that might not feel as painful is to remove the heading role? Is that needed? I am not really sure this text is a heading, even if it is big or was originally inside `<h2>`, which IMO was simply a mistake.

blog/templates/blog/_sidebar_with_categories_and_release.html
```diff
@@ -27,7 +27,7 @@ <h3 class="release-preview__title">{{ page.release_title }}</h3>
 	{% if page.get_current_release %}
 		{% with current_release=page.get_current_release %}
 			<h4 class="release-preview__tag">SecureDrop {{ current_release.tag_name }}</h4>
-			<div class="release-preview__date">{{ current_release.date|date:"E j, Y"}}
+			<div class="release-preview__date">{{ current_release.date|date:"E j, Y"}}</div>
```
chigby, line 30: :sweat_smile:

chigby (approved):
These changes look good to me!

common/templates/common/_footer.html
SaptakS: Yeah. That's a good point. I just tried to keep the same behaviour. But definitely if it's just a big sized text visually, and not a heading by content flow, then we can definitely remove heading and aria-level

Issue #865, opened 2021-06-23
Related label information for fields


SaptakS:
There are certain form fields which don't have labels related to them or have broken labels with missing id to associate with:
- [x] In https://securedrop.org/directory/ , the `<select>` dropdown needs to have an `id=aria-topic` for the label tag to be related with it.
- [x] In https://securedrop.org (and other parts of the website), the search field needs to have a label associated with it.
- [ ] ~~#887~~

chigby:
The third item on this list I think is ultimately coming from within Django. The wagtail form builder relies on Django to render checkbox list form fields and these are done without proper grouping tags.

Looks like it might be being [worked on](https://code.djangoproject.com/ticket/32338), though?

SaptakS:
I think we can wait on the django PR to be merged to resolve the third point for now. Depending on how much time it takes we can decide if we want to address it on our side.

SaptakS:
Tracking the third point as a [separate issue](https://github.com/freedomofpress/securedrop.org/issues/887). Closing this.

PR #886, opened 2021-08-16
Label accessibility

chigby:
This PR fixes the first two tasks in #865:

1. Adds a missing `id` attribute on the directory page for the Topic dropdown
2. Adds a screen-reader only label for the search box on the Search page.


SaptakS (approved):
LGTM

PR #893, opened 2021-09-13
Updates sqlparse and axios

SaptakS:


chigby (approved):
Looks good to me! Technically I don't think version 0.3.0 of sqlparse was affected by the vulnerability but it's good to keep these things up to date.

SaptakS:
True but safety checks were failing, so I just updated anyways.

PR #895, opened 2021-09-20
Adds a skip link to every page and adds context for links

### LINKED ISSUE #872: We should add a keyboard functional skip link to skip the navigation
We should add a keyboard functional skip link to skip the navigation and jump directly to the main content. Few things to keep in mind:

- The skip link should be selectable by keyboard, screen readers, etc.
- The skip link should be the first focusable element on the page
- It should be the first element inside the <body> tag
- Preferably always be visible, but at least should be visible on keyboard focus.

WCAG [2.4.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-skip.html)

### LINKED ISSUE #873: Add more context for the different links in the website.
Add more context for the different links in the website. For example, instead of just writing Release notes & Documentation & Read more, make it clear what the context of the link is. Also, same texts should have same link, and same link shouldn't have different text.

Solution:
we can add aria-label to describe the link  without needing to change the actual text of the link.

`<a href="https://docs.securedrop.org" aria-label="Documentations about SecureDrop">Documentation</a>`

So the minimum things to do while fixing through the links are:
- [ ] Add `aria-label` with proper context about the link if the link text is generic, or it is same all over. For example "View Directory.." link, or Read More links, etc.
- [ ] If same links have different text, add `aria-label` for the links which are same describing what the link is pointing to.

WCAG [2.4.4](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-refs.html), [2.4.9](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-link.html)

SaptakS:
Fixes #873 
Fixes #872 

This PR adds a "Skip to content" link to every page that takes the user directly to the page-title of the page. The link is not visible unless someone uses tab to navigate and focuses on that link (which is always the very first link in the page). This is how it looks when focused:
Screenshot from 2021-09-20 19-58-57

Also, in this PR, I have added aria-label to links which have generic text like "Read more", "View in directory" to add more context. It doesn't change any visuals, but gives more context to screen reader and other assistive tech users.

### DIFF
--- a/marketing/templates/marketing/_features_list.html
+++ b/marketing/templates/marketing/_features_list.html
@@ -3,7 +3,7 @@
 <div class="features-list">
 	{% for feature in features %}
 		<div class="features-list__item-wrapper">
-			<a class="features-list__item" href="{% pageurl feature.feature %}">
+			<a class="features-list__item" href="{% pageurl feature.feature %}" arial-label="{% trans 'Know more about the feature: ' %}{{ feature.feature.teaser_title }}">
 				<div class="features-list__item-content">
 					{% if feature.feature.icon %}
 						{% image feature.feature.icon max-200x200 class='features-list__item-image' %}
--- a/marketing/templates/marketing/_features_list.html
+++ b/marketing/templates/marketing/_features_list.html
@@ -4,10 +4,10 @@
 <div class="features-list">
 	{% for feature in features %}
 		<div class="features-list__item-wrapper">
-			<a class="features-list__item" href="{% pageurl feature.feature %}" arial-label="{% trans 'Know more about the feature: ' %}{{ feature.feature.teaser_title }}">
+			<a class="features-list__item" href="{% pageurl feature.feature %}">
 				<div class="features-list__item-content">
 					{% if feature.feature.icon %}
-						{% image feature.feature.icon max-200x200 class='features-list__item-image' %}
+						{% image feature.feature.icon max-200x200 class='features-list__item-image' role='presentation' %}
 					{% endif %}
 					<div class="features-list__item-text">
 						<h2 class="features-list__item-heading">{{ feature.feature.teaser_title }}</h2>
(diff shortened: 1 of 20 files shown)


marketing/templates/marketing/_features_list.html
```diff
@@ -1,9 +1,10 @@
 {% load wagtailcore_tags wagtailimages_tags %}
+{% load i18n %}
 
 <div class="features-list">
 	{% for feature in features %}
 		<div class="features-list__item-wrapper">
-			<a class="features-list__item" href="{% pageurl feature.feature %}">
+			<a class="features-list__item" href="{% pageurl feature.feature %}" arial-label="{% trans 'Know more about the feature: ' %}{{ feature.feature.teaser_title }}">
```
harrislapiroff: Do I understand correctly that aria-label prevents text inside the element from being read? I don't think we actually want that here since the text is meaningful.

harrislapiroff (changes requested):
Design changes look fine, but one minor comment on one of the other accessibility changes. Looks really great otherwise!

marketing/templates/marketing/_features_list.html
SaptakS: @harrislapiroff that is correct. This is what the screen reader reads out:

Screenshot 2021-09-21 at 1 44 50 AM

If that looks good to you, can remove aria-label from this.

harrislapiroff: Yeah, I think that's preferable to the label. You could mark the image as presentation imo, but I'll leave it up to you how much detail screenreader users want. With that fixed, this PR is good to go! I'll approve it provisionally.

harrislapiroff (approved):

SaptakS:
Weird that the build failed in this PR 🤔

harrislapiroff:
I wonder if it's a pipenv change. I think we do install the latest on every CI run. I'll rerun to see.

harrislapiroff:
Er, I guess it would be pip not pipenv

SaptakS:
@chigby @harrislapiroff passing now!

PR #896, opened 2021-09-27
Upgrade version script, add /health/version

maeve-fpf:
This adds a version endpoint, equivalent to https://github.com/freedomofpress/securethenews/pull/373. We have the same copy of the script (easier to read, new more-understandable branch output, but does not address the running at build time question), and a test. Basically nothing different.


SaptakS:
Seems like there are some linting issues.

SaptakS (approved):

PR #902, opened 2021-11-08
Upgrade npm and pip dependencies

SaptakS:
Refs https://github.com/freedomofpress/fpf-www-projects/issues/231

- Updates npm packages without pinning
- Updates makefile to be more consistent with other repos and to add pip upgrade commands
- Updates all pip dependencies
- Makes wagtail autocomplete related changes
- Updates sass deprecations (**NOTE: it still throws deprecation warnings because some of the dependencies have not updated their sass code**.)
- Pins psycopg2 to below 2.9 since it doesn't work with django 2.2
- Updates breaking factory imports
- Updates url tests in CSS scraper to use raw URLs instead of string URLs (**NOTE: This is a [change in the CSS tokenization standards](https://www.w3.org/TR/css-syntax-3/#changes-CR-20140220) but there are still many CSS codes, including me many times, when we use string URLs instead of raw URLs. ~~So we might need to add tests so it can scrape and test string URLs too. We will need to check the FunctionBlock node and see if the function name was "url" and then append the string value~~**)
- Updates url extract code for CSS to extract string URLs as well.


SaptakS:
PS: The deprecation warning for sass slash div I was getting for a dependency was from a hamburgers npm package which doesn't seem like its maintained anymore and hence the PR related to that is not merged: https://github.com/jonsuh/hamburgers/pull/105

I think we can move away from it maybe since it's our direct dependency and not a dependency of a dependency?

Makefile
```diff
@@ -75,6 +75,22 @@ upgrade-pip-dev: ## Uses pip-compile to update dev-requirements.txt for upgradin
...
+.PHONY: pip-upgrade
+pip-upgrade: ## Uses pip-compile to update all requirements that are not pinned
+# in requirements.in
+	docker run -v "$(DIR):/code" -w /code -it python:3.9-slim \
+		bash -c 'apt-get update && apt-get install gcc libpq-dev -y && \
+    pip install pip-tools && \
+		pip-compile --generate-hashes --no-header --allow-unsafe --upgrade --output-file requirements.txt requirements.in && \
+		pip-compile --generate-hashes --no-header --allow-unsafe --upgrade --output-file dev-requirements.txt dev-requirements.in'
+
+.PHONY: pip-dev-ugrade
```
chigby: Typo: should be "upgrade" here

README.rst
```diff
@@ -167,15 +167,15 @@ There are separate commands to upgrade a package without changing the ``requirem
...
+    make pip-update PACKAGE=package-name
 
 will update the package named ``package-name`` to the latest version allowed by the constraints in ``requirements.in`` and compile a new ``dev-requirements.txt`` and ``requirements.txt`` based on that version.
 
 If the package appears only in ``dev-requirements.in``, then you must use this command:
 
 .. code:: bash
 
-    make upgrade-pip-dev PACKAGE=package-name
+    make pip-dev-update PACKAGE=package-name
```
chigby, line 178: :tada:

chigby (approved):
This is looking really great, especially in the area of CSS asset processing. Very glad to have that particular construction now covered. 

I did make one small typo request but it's definitely not a big deal. I can merge one that's fixed.

PR #904, opened 2021-11-24
Bump docker-compose postgres image to 13

maeve-fpf:
Per web sync, let's upgrade this to match our infrastructure. This one is a little different due to the slightly customized image that Mike created a while back. All it does is add a script to enable `hstore`, which seems like something that app migrations can handle. Maybe we can remove it?


chigby:
I'm pretty sure it can be done with migrations.. want me to add that to this PR?

chigby (approved):
Looks good, and now CI is passing due to the addition of a `version` argument to the `setup_remote_docker` command.

PR #915, opened 2022-01-24
Update python and npm dependencies

SaptakS:
This PR updates django, wagtail, pillow, ipython and npm dependencies like follow-redirects.

Updating to django==3.2.11 and wagtail==2.15.2 needed certain changes:
- staticfiles template tags had been deprecated in django 2.x and completely removed in django 3. Now one needs to just use `{% load static %}` instead. Locally the files seemed to be loading correctly, but we need to verify the same in staging in case some changes is needed in the settings file
- when using `annotate` in queries, where there are mixed model types, need to provide an additional `output_field` to specify the model type we want the final output field to have.
-  `NullBooleanField` has been deprecated and needed to be replaced with `BooleanField(null=True)`
- Added settings for wagtail comments to be disabled to be consistent with PFT settings along with updating DEFAULT_AUTO_FIELD and search backends.
- Django tests now has different meaning for `-k` flag and needed to be replaced with `--keepdb`


chigby (approved):
Everything looks good to me here, probably worth some testing on staging.

PR #947, opened 2022-08-03
Updates django for security vulnerability

SaptakS:
https://www.djangoproject.com/weblog/2022/aug/03/security-releases/ it's marked by the django team as severe vulnerability. Though we don't use FileResponse anywhere I think and even if it's used indirectly, we don't send back files as response as such anywhere. So might not be super urgent, but still good to have it deployed soon.


chigby (approved):
Security changes look good here

PR #955, opened 2022-08-15
Updates node version in ProdDjangoDockerfile

SaptakS:
I realized that I didn't update the node version in the ProdDjangoDockerfile. Hence this PR.


chigby (approved):
Good catch, merging this.

PR #977, opened 2022-11-17
Wagtail 4.1

### LINKED ISSUE #974: Upgrade to Wagtail 4.1

chigby:
Fixes #974

See individual commits for rationale behind changes. Most are following through with wagtail's upgrade instructions.


SaptakS (approved):
I have gone through the changelog and verified that all the changes mentioned in the changelog have been addressed. Also checked by running locally and seems like everything is working. The live preview is a bit slow in my local system, and takes some time to update, but that can be just my system. Also, I don't think that's a blocked.

PR #978, opened 2022-11-22
Adds CSP rules explicitly for directives not covered by default-src

SaptakS:
Refs #https://github.com/freedomofpress/infrastructure/issues/3372

### DIFF
--- a/securedrop/settings/base.py
+++ b/securedrop/settings/base.py
@@ -279,6 +279,11 @@ SOCIALACCOUNT_EMAIL_REQUIRED = True
 # #11 needed for wagtail admin
 
 CSP_DEFAULT_SRC = ("'self'",)
+CSP_BASE_URI = ("'self'",)
+CSP_FORM_ACTION = ("'self'",)
+CSP_FRAME_ANCESTORS = ("'self'",)
+CSP_PLUGIN_TYPES = ("'self'",)
+CSP_SANDBOX = ("'self'",)
 CSP_SCRIPT_SRC = (
     "'self'",
     "'unsafe-eval'",


securedrop/settings/base.py
```diff
@@ -279,6 +279,11 @@
 # #11 needed for wagtail admin
 
 CSP_DEFAULT_SRC = ("'self'",)
+CSP_BASE_URI = ("'self'",)
+CSP_FORM_ACTION = ("'self'",)
+CSP_FRAME_ANCESTORS = ("'self'",)
```
chigby, line 284: Is this supposed to be `'none'`? That's my reading of the [original issue](https://github.com/freedomofpress/infrastructure/issues/3372) but maybe I'm wrong.

mig5: It depends on your site's functionality. If you don't specify anything at all for those other attributes, it's the same as saying 'allow from anywhere' because the `default-src` is not respected for these special case attributes.

So setting anything to 'none' might break functionality that has transparently been allowed til now. But conversely, 'self' might be too broad a permission if you really don't intend on the site needing it (but still better than the 'allow from anywhere' that is caused by not specifying it at all).

If your app needs to submit forms whose 'action' (or which perform any redirect after the action) return the user back to the same app, then 'self' is right. If it has no forms, 'none' is the most secure option. If submitting a form takes you to any other site as a redirect, you need to have that site listed in the form-action.

If it needs to be able to be embedded as an iframe in other sites, then frame-ancestors should not be 'self' but something else (if you don't want the site to be embedded as an iframe by any other site, then set frame-ancestors to 'none'. Basically you want this value to be the same as the (defunct) X-Frame-Options header used on the site.

Base URI I'm less clear on, it is a question of whether the site uses `<base>` HTML tags anywhere. If you don't think you have any, then maybe 'none' is the best option.

SaptakS: exactly what @mig5 said. I just put self instead of none to kind of future-proof it at least to changes that we make at our own code (i.e. self). So "self" is little safer but strict enough (since it still blocks any external src), but "none" would be stricter. That was my thinking.

chigby: Gotcha, that makes sense to me.

chigby (approved):
Everything seems good here, I don't think we're quite at the confidence level to put "none" on anything yet, but we can definitely revisit that in the future.

securedrop/settings/base.py
mig5: Just be aware especially for the form-action directive, browsers treat the CSP differently. Chrome is very strict.

Let's say you had a form action of /send (a route on the same site)

Then after /send processes some action, it did a 302 redirect to some-other-site.tld/thanks

Chrome will block the submission of the form if some-other-site.tld/thanks is not in the form-action statement in the CSP, even if technically, the 'form action' is /send.

Firefox on the other hand considers it that you only need to specify 'self' if the 'action' was a relative route on the same site.

I don't think that's going to be an issue on the SDO site (though I believe there's a contact form, so thought I'd mention it), I just wanted to pass along some hard lessons learned :) Feel free to hit me up on other channels if you have any questions!

PR #979, opened 2022-11-23
Removes sandbox directive from CSP rules

SaptakS:
The sandbox directive doesn't support the "self" value. It allows a tuple of values of allow-* related to things that are allowed. Keeping with the logic of being safe, making a tuple might be less flexible for us and might need much more thorough testing. Hence removing the directive for now. As far as my understanding, this isn't really that risky of a directive to skip.


chigby (approved):
This checks out in my testing, so let's merge it!

PR #993, opened 2023-01-24
Docker and Readme Update

chigby:
Started as part of https://github.com/freedomofpress/fpf-www-projects/issues/297 and if we like this technique I can port it to our other repositories.

This pull request:

1. Updates the documentation for the initial docker setup to use the correct form of the "docker compose" command and generally be more useful and accurate.
2. Updates the development dockerfile to take advantage of multi-stage builds, particularly for the frequently used command "compile-pip-dependencies". See commit message for more details. I didn't change the other pip commands since I'm not sure how/if they are being really used by anyone. Maybe they should be updated too, or made nicer? Not sure.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,7 +36,7 @@ To get started, run:
 .. code:: bash
 
     make dev-init  # one-time command
-    docker-compose up  # long-running process to run application server, every time
+    docker compose up  # long-running process to run application server, every time
 
     # In a separate shell:
     docker-compose exec django ./manage.py createdevdata  # one-time command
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -23,30 +23,42 @@ Table of Contents
...
 
 .. code:: bash
 
     make dev-init  # one-time command
-    docker-compose up  # long-running process to run application server, every time
+    docker compose up  # long-running process to run application server, every time
 
     # In a separate shell:
-    docker-compose exec django ./manage.py createdevdata  # one-time command
+    make dev-createdevdata  # one-time command
```
harrislapiroff, line 52: I don't feel super strongly about this, but I do actually kinda like encouraging devs to use and get used to `docker compose` commands rather than abstracting them away and this doesn't seem like an overly complicated one that requires a shortcut. (As discussed in web sync, the command already existed and some of us use it and some don't.)

harrislapiroff: After discussion, I'm fine with leaving this since we have the command. If and when we migrate to poetry I'd like to see us use fewer shortcuts and more explanations of how to use our tooling directly :)

harrislapiroff (approved):
LGTM

PR #999, opened 2023-02-15
Silence alerts for alerts related to cryptography <= 39.0.0

chigby:
We are not at all in a position to upgrade cryptography right now.

If this passes the local safety check, I will be following up with PRs to centralize the ignore list, as in https://github.com/freedomofpress/fpf-www-projects/pull/316


SaptakS:
I think once https://github.com/freedomofpress/securedrop.org/pull/1000 is merged and this PR is rebased on it, the safety should pass. But right now it's a circular safety-check fail. :P

SaptakS (approved):

PR #1005, opened 2023-03-08
Npm dependency upgrades

### LINKED ISSUE #952: Update npm dependencies
Update all the JS dependencies to their latest version. One can check the outdated npm dependencies using: `npm outdated`. If it's a minor release, one can just update using `npm updata <package-name>`. If it's a major release update, then one needs to update the package.json and then run `npm install` (or `npm i --package-lock-only` if you don't want to install)

chigby:
Fixes #952 

Upgrades the packages we're still using and removes a number of them that we are not.


package.json
```diff
@@ -17,34 +17,25 @@
     "@babel/preset-env": "^7.18.10",
     "@babel/preset-react": "^7.18.6",
     "autoprefixer": "^10.4.2",
-    "babel-loader": "^8.2.5",
+    "babel-loader": "^9.1.2",
     "babel-plugin-add-module-exports": "^1.0.4",
     "css-loader": "^6.7.1",
     "mini-css-extract-plugin": "^2.6.1",
-    "npm-audit-plus": "^0.1.7",
```
harrislapiroff, line 24: The switch to `npx` above made me think we could invoke it in such a way to obviate the need to install this via dependencies at all! But we'd probably lose some performance by not baking the dependency into the docker image... I guess better to leave as is.

harrislapiroff (approved):
LGTM, but I'm gonna hold off merging till we've resolved deploy issues

PR #1010, opened 2023-03-16
Expand the Directory Instructions Template

nathandyer:
This PR expands the directory instructions to provide some additional clarity to potential sources.

The primary change is that it makes it clear that seeing a different onion address in the URL bar inside of Tor Browser for securedrop.org does not constitute an onion address mismatch with an organization's Source Interface.

It also provides some slight tweaks to make the workflow a bit clearer.

If possible, I would appreciate @eloquence's review of the content, in addition to the standard review procedure by the web team. Thanks!

### DIFF
--- a/directory/templates/directory/_instance_instructions.html
+++ b/directory/templates/directory/_instance_instructions.html
@@ -3,21 +3,42 @@
 <div class="instance-instructions">
 	<h1 class="instance-instructions__title">{% trans "Planning to submit information?" %}</h1>
 	<p>
+		{% blocktrans with organization=page.title %}
+			The button below will take you to {{ organization }}’s landing page,
+			which will provide more details about how to submit files to their SecureDrop instance.
+		{% endblocktrans %}
+	</p>
+	<p>
+		Because the landing page is linked in our directory, it was found to be in compliance with
+		our strict security requirements.
+	<p>
 		{% blocktrans with organization=page.title onion=page.full_onion_address %}
-			Be sure that the .onion address provided on {{ organization }}’s landing
-			page below matches
+			As part of the submission process, we have confirmed {{ organization }}’s SecureDrop to be
+			available at the following .onion addresses:
+			<br />
+			<br />
 			<span class="instance-instructions__onion-address">{{ onion }}</span>
 		{% endblocktrans %}
 		{% if page.onion_name %}
 			{% blocktrans with onion_name=page.onion_name %}
-				or <span class="instance-instructions__onion-address">{{ onion_name }}</span>
+				<br />
+				<span class="instance-instructions__onion-address">{{ onion_name }}</span>
+				<br />
+				<br />
 			{% endblocktrans %}
 		{% endif %}
+	</p>
+	<p>
 		{% blocktrans %}
-			before continuing. This helps ensure you are visiting a legitimate
-			SecureDrop instance. See our
+			Please ensure the .onion address you are directed to visit matches an address above.
+			This helps ensure you are visiting a legitimate SecureDrop instance. See our
 			<a href="https://docs.securedrop.org/en/stable/source.html">Source Guide</a>
 			for more information on SecureDrop usage and risks.
+	</p>
+	<p>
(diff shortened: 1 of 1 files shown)


eloquence (changes requested):
Thanks for your work on this, Nathan! Here's what this change looks like when I build it locally:

image

Compared with the current version:

image

I think we have a bit of a verbosity issue (even in the old version), which can lead to folks just glancing over the instructions and missing the most essential bits.

I'll put together a GDoc where we can workshop this a bit together before revising the PR.

directory/templates/directory/_instance_instructions.html
```diff
@@ -2,22 +2,43 @@
...
-			SecureDrop instance. See our
+			Please ensure the .onion address you are directed to visit matches an address above.
+			This helps ensure you are visiting a legitimate SecureDrop instance. See our
 			<a href="https://docs.securedrop.org/en/stable/source.html">Source Guide</a>
 			for more information on SecureDrop usage and risks.
+	</p>
+	<p>
+			Note that this .onion address will be different than the .onion address you
+			will see in your browser's address bar above, if you are visiting this page
+			using the Tor Browser.
```
eloquence: My understanding is that Tor Browser, by default, only shows the .onion as ".onion available" - you can turn on a preference to automatically go to the .onion.

eloquence:
I've started [this doc](https://docs.google.com/document/d/1C4n5UeW_eQRVc0wnUUrVJGLPQfkrqUm8PwsNLfp1YTo/edit) to workshop the instructions further :)

nathandyer:
Thank you for putting together that doc so we can workshop the instructions, @eloquence! I agree that both the original and modified text here in the PR is too verbose, and really like the changes you've put together so far. I made one suggested tweak to try and capture the nuance about the .onion address mismatches (clarifying to be on the lookout for the ones on the landing page and comparing them to the ones listed in the directory guide). Please feel free to tweak that further or make an alternate suggestion. Otherwise, I'm in favor of what you have put together so far, and will be glad to update the PR accordingly.

nathandyer:
I updated this PR with the latest text from our workshopped version. I was a bit unsure about the logic related to the GPG email preferences, so if I need to make tweaks with that (or any of it), please let me know. Thanks!

directory/templates/directory/_instance_instructions.html
```diff
@@ -1,37 +1,41 @@
...
 		{% endif %}
-		{% blocktrans %}
-			before continuing. This helps ensure you are visiting a legitimate
-			SecureDrop instance. See our
-			<a href="https://docs.securedrop.org/en/stable/source.html">Source Guide</a>
-			for more information on SecureDrop usage and risks.
+	</p>
+	<p>
+		{% blocktrans with organization=page.title %}
+			If the .onion address information on {{ organization }}’s web page
```
eloquence: Suggestion after seeing it in context: Maybe we should use "landing page" instead of "web page" here since we're using in the link description and the warning copy. (It's our internal jargon, so I almost would prefer to get rid of it completely, but if we're using it, using it consistently may help with pattern-matching.)

eloquence (approved):

eloquence:
Left one suggestion but wording looks good from my end either way! Leaving for web team to merge in case I'm missing anything, but I don't see an issue with the logic -- the basic idea is that if a page like https://securedrop.org/report-an-error/ is configured (which it is in prod), we'll link to that instead of the email address.

directory/templates/directory/_instance_instructions.html
nathandyer: Thanks @eloquence, I agree - I've updated this to use the term "landing page" instead.

eloquence:
Rebased; flagged to SD team in case of any concerns about language update, so keeping "in progress" for now.

eloquence:
3 thumbs-ups from SD team members so I think this is good for a quick final pass from @freedomofpress/fpf-web-developers  :)

chigby (approved):
Everything seems to look good from my perspective.

PR #1019, opened 2023-04-25
Updates postcss-loader to remove vuln

SaptakS:
The vulnerability is actually in the yaml package, which is a dependency of cosmiconfig, which is a dependency of postcss-loader. However updating postcss-loader, updates cosmiconfig, and the cosmiconfig latest version uses js-yaml instead of yaml. So I have just updated the entire dependency chain to hopefully resolve the vulnerability.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1753,6 +1753,30 @@
         "node": ">=6.9.0"
       }
     },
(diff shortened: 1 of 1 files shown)


chigby (approved):
Looks good to me (and the audit script)

PR #1033, opened 2023-07-03
Updates django to 3.2.20

SaptakS:
You can read about the vuln here https://www.djangoproject.com/weblog/2023/jul/03/security-releases/


harrislapiroff (approved):
LGTM

PR #1046, opened 2023-08-23
Leave wagtail search backend on default configuration if possible

chigby:
## Description

The settings variable `WAGTAILSEARCH_BACKENDS` is assigned to the database backend in the base settings file, but in this production file we are checking to see if ElasticSearch is enabled.  If it is (not the case, IMO) then we configure the backend, and otherwise reset the backends setting to `{}` -- I'm not sure how this actually affects the search feature on the sight, but there's no reason to do it.  We want the DB search backend in the default case, so let's leave it that way.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [x] Config changes

## Testing

This ideally either changes nothing, or fixes the search if it's broken. Right now, I'm not getting any results for any searches in the wagtail admin. I recommend testing the admin page/document search on staging. It should not raise any errors, and if it actually finds results, then that's a bonus!

### Post-deployment actions

None.

### General checks

### If it's a major change

No frontend changes.

### DIFF
--- a/securedrop/settings/production.py
+++ b/securedrop/settings/production.py
@@ -168,9 +168,7 @@ else:
 try:
     es_host = os.environ.get('DJANGO_ES_HOST', 'disable')
 
-    if es_host == 'disable':
-        WAGTAILSEARCH_BACKENDS = {}
-    else:
+    if es_host != 'disable':
         WAGTAILSEARCH_BACKENDS = {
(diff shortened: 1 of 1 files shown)


SaptakS (approved):
Makes sense to me.

PR #1051, opened 2023-09-05
Remove pshtt

### LINKED ISSUE #1008: [Spike] Remove pshtt to see what breaks
# Description

Before we update https://github.com/freedomofpress/pshtt to use the latest `sslyze` package (we're currently on 3.0.0, and you can see that there have been quite a few changes since then: https://github.com/nabla-c0d3/sslyze/compare/3.0.0..5.1.2), let's evaluate whether or not we want to continue maintaining https://github.com/freedomofpress/pshtt. Here are some open questions to get us started:

* Is `pshtt` only used for the Scan Results page/ landing page alerts feature?
* How do we use this feature in practice? 
* Should we reconsider moving back to the original repo instead of maintaining our own fork and then just making PRs to update dependencies?
* Are there alternative tools that are better maintained?
* Any other considerations?

# Background

We maintain a fork of `pshtt` (for background as to why and when, see https://github.com/freedomofpress/fpf-www-projects/issues/150). Four days ago, `sslyze 5.1.2` was released with dependency upgrades, which included an upgrade to `cryptography` (see https://github.com/nabla-c0d3/sslyze/releases/latest). This means we can update our fork, which will subsequently mitigate the cryptography alerts that we've been ignoring (note: the web team determined that it was safe to ignore these since we only use `pshtt` to scan 3rd party sites). @harrislapiroff - please chime in if I missed any relevant background information that might be helpful.

chigby:
## Description

Fixes #1008

Changes proposed in this pull request: pshtt has been removed.

These scan result fields were dependent on data from pshtt:

* `live` -- I've changed this to be `False` if the request to the directory landing page raises a `requests` exception, and `True` otherwise. This should duplicate the existing functionality, which mapped `live` to the pshtt data's `Live` field. The purpose of this field is to determine if the landing page is reachable. There is also a separate field that records whether the response is a 200 status, so we have some redundancy here as well.
* `http2` -- We have logic to determine if a remote server supports the http2 protocol. We were using this logic on the `Canonical URL` field from pshtt. I have changed it to the [`url` field from the `requests` response](https://requests.readthedocs.io/en/latest/api/#requests.Response.url), which is the final URL that was reached by our request after all redirects. This should have the same behavior as before, except in cases where pshtt was inferring "canonicity" between www and non-www versions of the same domain. I do not think this checking is hugely important, especially for checking http2 support, but if it is, we might want to rethink this.
* `forces_https`
* `hsts`
* `hsts_max_age`
* `hsts_entire_domain`
* `hsts_preloaded` -- All of these HTTPS fields are being populated by pshtt and we are _not_ using them in a warning anywhere. I have changed the code to keep them as `None` for future scan results.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Testing

Try scanning an existing directory or a URL. The scan should complete without errors. The https-related fields should all be `None` and the `live` and `http2` checks should complete as they usually have.

Also, here's how we can verify that the warnings feature of the scanner is still working. Below are the 5 warnings we support and URLs that would trigger those warnings. These can be tested on staging (once this code is merged) or locally by opening a directory entry page and updating the landing page URL to one of the ones below. Note that in order to see the moderate warnings, the third party assets warning must be ignored, see description.

* Severe: no_third_party_assets: https://proton.me/tor -- Note: all of these links will receive this warning, which will override moderate-level warnings. To see the moderate warnings on the URLs on this list, "Use of analytics or third party assets" must be checked on the "warnings ignored" list in the Settings tab for that directory entry.
* Severe: unreachable_landing_page: https://google.com/not-there
* safe_onion_address: https://proton.me/tor
* referrer_policy_set_to_no_referrer: https://pypi.org/
* subdomain: https://docs.djangoproject.com/en/4.2/

### Post-deployment actions

None.

### General checks

No changes.

No changes.

- [x] Verify that the directory scan result page in admin interface loads as expected

### If it's a major change

No frontend changes.


eloquence:
The approach in this PR makes sense to me in principle.

I would suggest:
- We update the test plan to add some additional steps to confirm that existing warning functionality (the warnings displayed on directory entries when certain checks fail) works as expected.
- We track full removal of unnecessary fields in a follow-up issue (if we can live with this config for a few months, I don't see a compelling reason to keep the historical scan data around).

Once it's merged, please give me a heads up before deployment to prod, and I can make sure we also verify it on staging one more time.

dev-requirements.txt
```diff
@@ -950,9 +772,6 @@ protobuf==4.22.1 \
     #   -r requirements.txt
     #   google-api-core
     #   googleapis-common-protos
-pshtt @ https://github.com/freedomofpress/pshtt/zipball/65596ff08fa7bf0357b5af63da73e2dead91c304 \
-    --hash=sha256:56e10b98409223c2800c25268822468579a7dad0faebc8ed31f9d12ef7b80381
-    # via -r requirements.txt
```
SaptakS, line 955: ❤️

scanner/scanner.py
```diff
@@ -350,14 +329,6 @@ def validate_hsts_max_age(max_age):
         return False
```
SaptakS: I guess we can remove `validate_hsts_max_age` function as well.

SaptakS (changes requested):
I have left a small inline comment about removing an unused function.

SaptakS:
One thing I am wondering while reviewing this PR is if we are using the scan data just for the warnings shown in the directory listing, and nothing else, then should we remove the other scans as well? There are only like 5-6 attributes that are being used for showing warnings, but we have like  30+ attributes that we are computing during the scan. Though this is not related to the PR, but probably something to discuss.
cc @harrislapiroff @eloquence

harrislapiroff:
We want to be sure to have some time for @eloquence to test this on staging before deploy.

scanner/scanner.py
chigby: Good idea here. I've removed the function.

chigby:
Ready for re-review.

SaptakS (approved):
Looks good to me.

SaptakS:
@harrislapiroff should I hold off on merging this? In case @eloquence doesn't have time to test this in staging before this week's deploy.

eloquence:
I've also double-checked the warning behavior. Warnings/overrides work as expected. No concerns moving forward with a deploy.

PR #1050, opened 2023-09-05
Update django to 3.2.21, gitpython to 3.1.34

chigby:
## Description

Fixes vulnerablities described here:

- django: https://docs.djangoproject.com/en/4.2/releases/3.2.21/
- gitpython: https://data.safetycli.com/vulnerabilities/CVE-2023-41040/60841/

Changes proposed in this pull request:

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Vulnerabilities update
- [ ] Config changes
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires an admin update after deploy
- [ ] Includes a database migration removing  or renaming a field

## Testing

Not sure if testing needed -- this change should fully be compatible with our codebase.

### Post-deployment actions

None.

### General checks

- [x] Linting and tests pass locally
- [x] The website and the changes are functional in Tor Browser
- [x] There is no conflicting migrations
- [x] Any CSP related changes required has been updated (check at least both firefox & chrome)
- [x] The changes are accessible using keyboard and screenreader

### If you made changes to directory listing:

N/A

### If you made changes to contact form

N/A

### If you made changes to scanner

N/A

### If it's a major change

Not a major change.

### If you made any frontend change

Not a frontend change.

PR #1052, opened 2023-09-07
Deploy 2023.09.07

chigby:
* Update gitpython and django (security fixes)


chigby:
Bypassing safety check to merge. The check is failing due to an unpatched CVE in gitpython in our dev dependencies, see https://github.com/freedomofpress/securedrop.org/pull/1050#issuecomment-1709021805

PR #1055, opened 2023-09-14
Only run `npm_audit` and `safety_check` jobs on develop

chigby:
This is intended to stop these from running on non-deployment pull requests, and only trigger them when we are merging from `develop` into `prod`.

## Description

Part of https://github.com/freedomofpress/fpf-www-projects/issues/360

Applies CircleCI filters to the `npm_audit` and `safety_check` jobs so they will only run on the `develop` branch. This is to stop them from running on all PRs except for ones we make to deploy the site from `develop` into `prod`.

I've also done a little housekeeping:
1. Updated the version of node used for the audit check
2. Tried to make the `build_dev` check easier to read by breaking it into multiple steps.
3. Moved `bandit` into the `build_dev` check so it still runs on every PR.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [x] Config changes
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Testing

Things to verify about this change:
1. This PR should not have the npm audit or safety check jobs appear on it.
4. After this is merged, the PR to deploy the site again _should_ have those jobs on it.

### Post-deployment actions

None.

### General checks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,14 +29,6 @@ jobs:
           name: Check Python dependencies for CVEs
           command: make safety
 
-      - run:
-          name: Update bandit to the latest version
-          command: pip install --upgrade bandit
-
-      - run:
-          name: Static code analysis for vulnerabilities
-          command: ./scripts/bandit
-
   npm_audit:
     docker:
       - image: cimg/node:20.6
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -29,17 +29,9 @@ jobs:
           name: Check Python dependencies for CVEs
           command: make safety
 
-      - run:
-          name: Update bandit to the latest version
-          command: pip install --upgrade bandit
```
dkaoster, line 34: does this no longer need to be installed / updated somewhere?

dkaoster (approved):

.circleci/config.yml
chigby: The story here is I've moved the place where we run `./scripts/bandit` to one of the following commands which is run within our docker compose cluster, and that setup installs bandit from our dev-requirements.txt

chigby: I.e. it's installed when `docker compose up` is run

dkaoster: sounds good

.circleci/config.yml
```diff
@@ -62,17 +54,30 @@ jobs:
...
+          name: Run flake8
+          command: make flake8
+
+      - run:
+          name: Validate migrations
+          command: make check-migrations
+
+      - run:
+          name: Static code analysis for vulnerabilities
+          command: docker compose exec django ./scripts/bandit
```
SaptakS: My only small comment would be that we can create a `make` command for this as well (since that's the practise with all the other checks so feels consistent to me). @chigby let me know what you think.

SaptakS (approved):

PR #1065, opened 2023-10-03
Update urllib3

chigby:
## Description

Urllib3 updates fixes this vulnerability:

* https://github.com/advisories/GHSA-v845-jxx5-vc9f

I also removed it from requirements.in since technically it is only a requirement of `requests`, not of our own code. Also setuptools was updated as part of pip-compile's update process, I suppose. I don't think that's hurting anything.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Vulnerabilities update
- [ ] Config changes


harrislapiroff (approved):
LGTM

PR #1080, opened 2023-11-16
Adds robots.txt and sitemap.xml

### LINKED ISSUE #1075: Add robots.txt
As with https://pressfreedomtracker.us/ we should have a robots.txt for this site defining expected crawler behavior and exclusions.

SaptakS:
## Description

Resolves #1075.

Adds a robots.txt which allows all the URL to all bots (or rather disallows nothing). Also adds a sitemap.xml and adds the sitemap link to robots.txt.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- Check /robots.txt shows the content that it should.
- Check /sitemap.xml shows the sitemap correctly.

PS: Unlike in PFT, I have not add any cache purging for sitemap specifically since it seems like whenever a new page is created or deleted in securedrop.org, all cache is purged. Let me know what you think @chigby

### DIFF
--- a/common/templatetags/common_tags.py
+++ b/common/templatetags/common_tags.py
@@ -72,3 +72,10 @@ def get_site_name():
 @register.filter
 def document_view_url(value):
     return reverse('view_document', args=[value.id, value.filename])
+
+
+@register.simple_tag(takes_context=True)
+def get_absolute_url(context, view_name, *args, **kwargs):
+    return context['request'].build_absolute_uri(
+        reverse(view_name, args=args, kwargs=kwargs)
+    )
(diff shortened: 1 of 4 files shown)


chigby:
@SaptakS it does indeed look to me like we're purging everything whenever a page is published or deleted, according to this code:

https://github.com/freedomofpress/securedrop.org/blob/develop/common/signals.py#L10-L16

So I think it's fine not to add anything special for the sitemap here.

chigby (approved):
This looks good to me. I checked locally for any 4xx/5xx responses from pages in the dev data that were on the site map and I didn't find any, so I think we're good to go.

PR #1085, opened 2023-12-07
Update to python 3.12.1,  django 4.2, wagtail 5.2

### LINKED ISSUE #1077: Upgrade to Wagtail 5.2.x, Django 4.2.x and Python 3.12
Wagtail has just [released](https://docs.wagtail.org/en/stable/releases/5.2.html) their new LTS that also allows us to move to the latest LTS of Django. Also, as discussed in meeting, we planned to move to the latest stable Python release that we can migrate to.

Deliverables:
- [ ] Upgrade to Wagtail 5.2.x
- [ ] Upgrade to Django 4.2.x
- [ ] Upgrade to Python 3.12 (or 3.11 in case all our dependencies don't yet support 3.12)

Once this is done, we would also need to let the editorials know about the changes in UI (and any new [features](https://wagtail.org/blog/curate-content-better-and-make-it-greener-in-wagtail-52/) they might be interested in)

SaptakS:
## Description

Fixes #1077.

- Updates to django 4.2 and make necessary code changes
- Updates to Wagtail 4.2 and make necessary code changes
- Updates to Python 3.12 and updates other python packages to support Python 3.12

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Dependency update
- [ ] Config changes

## Testing

It might be worth checking the changelogs to see if I missed anything. Other than that, try building the container locally with the new dependencies, check all the pages and admin section and tests work without throwing any errors or deprecation warning.


devops/docker/ProdDjangoDockerfile
```diff
@@ -20,8 +20,8 @@ RUN cd /src-files && ( npm install && npm run build )
 LABEL MAINTAINER="Freedom of the Press Foundation"
 LABEL APP="securedrop.org"
 
-# sha256 for python:3.9.18-slim-bookworm on linux/amd64 as of 2023-09-15
-FROM python:3.9-slim-bookworm@sha256:44b7f161ed03f85e96d423b9916cdc8cb0509fb970fd643bdbc9896d49e1cad0 AS python-base
+# sha256 for python:3.12.0-slim-bookworm on linux/amd64 as of 2023-12-06
+FROM python:3.12.0-slim-bookworm@sha256:8e216a21d8df597118b46f3fff477ed1c5c11be81531b6da87790a17851b7f1c AS python-base
```
chigby: I feel like we might as well add python 3.12.1 here (released last Thursday :sweat_smile:)

SaptakS: Yep! I did see that in the feeds.

chigby:
Noticing that `ModelAdmin` is [deprecated by wagtail](https://docs.wagtail.org/en/stable/releases/5.1.html#wagtail-contrib-modeladmin-is-deprecated) and will be removed in version 6. We're using that for managing scan results. 

Do you think this is something we want to address now, or wait until we have to do it?

SaptakS:
I think might as well address it now.

harrislapiroff:
Agree that seems like the right call

SaptakS:
@chigby I did migrate from ModelAdmin to SnippetViewSet, however something are not working like before:
- I am unable to update the name of the action button
- Can't set the create view correctly. I have added the correct class, but whatever class variable I set, it seems to somehow not be ever set or used (not even updating the page_title). Not sure what is happening there
- I haven't looked in depth for the filters, but right now it's just text searches.

If you could look into the create view class part of the code, would be helpful, since I seem to have hit a wall even after reading the source code.

SaptakS:
I have explained a bit on what I did in the commit comment, but just to summarize here, I tried to migrate to ModelViewSet like @harrislapiroff had suggested, but that kind of ended up having the same issues. However, after I started diving into the codebase of Wagtail, I found out that ModelViewSet has a class called `get_common_view_kwargs` which is used to set some view related kwargs including various URL names. I also found out that the `generic.IndexView` uses this URL names to set certain parameters which are then used by the template to check and create various URL parts in the page such as the create button, or the context menu or the edit page link on the page title in the listing. So I tried overriding that function and setting the `add_url_name`, `edit_url_name` and `delete_url_name` to `None`, and that seemed to remove the add, edit and delete options from everywhere in the page.

chigby:
Did we also lose the "Inspect" button on the scan result admin page? I'm not seeing how to get to that view (for a single result) on the index of all the results.

chigby:
Oh, it's in the "..." menu! Got it.

chigby (approved):
Everything here is looking good, and I think the modeladmin replacement is a great balance of keeping the previous UX while also using the new APIs. This is ready to ship, IMO.

PR #1087, opened 2023-12-12
Updates frame-src and connect-src with media subdomain

SaptakS:
## Description

It was reported that PDF embeds are not working in pages like https://securedrop.org/news/piloting-securedrop-workstation-qubes-os/. On testing, I found out that they worked fine in firefox but not in chrome. The reasoning in chrome seemed to be that the `media.` subdomain is missing from the `frame-src` CSP directive. However on investigating why it works in firefox, seems  like the redirects to our media subdomain is handled differently in the different browsers. In firefox, the initiator of the redirected link is considered to be the embed object, and hence the CSP directive checked is `object-src`, but in chrome, it still checks under the `frame-src` directive. We had already updated our setting before to add `media.` subdomains to our `object-src`.

So in this PR, I modified the code such that the `frame-src` and `connect-src` are both updated with the `DJANGO_CSP_OBJ_HOSTS` environment variable. In all the other web repo, we have been updating both `object-src` and `connect-src`. It seems like we are using pdf.js in the other repo which sends an XHRequest and hence falls under the `connect-src` directive. But in SDO, we are using Pdfoject, which creates an embed object whose redirect seems to be handled differently in different browser.

Hopefully that makes sense. We have to test if all of the above are true in staging, since locally I don't have a redirect for media.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

I don't think there is a good way to test it locally. So I think we might have to test it on staging and see if the `frame-src` CSP directive has `https://staging-media.securedrop.org`

### Post-deployment actions

In case this PR needs any admin changes or run a management command after deployment, mention it here:

### General checks

### DIFF
--- a/securedrop/settings/base.py
+++ b/securedrop/settings/base.py
@@ -292,11 +292,10 @@ CSP_STYLE_SRC_ATTR = (
     "'unsafe-hashes'",
     "'sha256-ZdHxw9eWtnxUb3mk6tBS+gIiVUPE3pGM470keHPDFlE='",
 )
-CSP_FRAME_SRC = ("'self'",)
-CSP_CONNECT_SRC = (
+CSP_CONNECT_SRC = [
     "'self'",
     "analytics.freedom.press",
-)
+]
 CSP_EXCLUDE_URL_PREFIXES = ("/admin", )
 
 # Need to be lists for now so that CSP configuration can add to them.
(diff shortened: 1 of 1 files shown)


chigby (approved):
I'm pretty sure I understand the issue as you've described it, and agree that we should go ahead and test this change on staging to see if it works. If not, well, then we will know more.

chigby:
[The staging site](https://staging.securedrop.org/news/piloting-securedrop-workstation-qubes-os/) does seem to work now with the PDF embedded on the blog page referenced in the PR description.

PR #1109, opened 2024-03-27
Moves default PR temlpate back to .github folder

SaptakS:
Related to https://github.com/freedomofpress/fpf-www-projects/issues/399. I think I misunderstood that the default PR template also needs to be in `.gtihub/PULL_REQUEST_TEMPLATE/` folder. It should still be just in `.github/` folder to be prefilled by default. That is my new understanding.

Once we merge this PR, I can test it in this repo, and if it works, I can make the same changes in the other repos.


harrislapiroff (approved):
Let’s try it!

SaptakS:
Just verified, seems like it is working.

PR #1126, opened 2024-05-29
Bulk scan error handling

### LINKED ISSUE #1125: Scan results not consistently updated
https://securedrop.org/admin/scanresult/ ordered by "Result last seen" currently shows 6 results for May 28 (today), 2 for earlier in May, and then drops off into March, February and January.

I would expect the whole directory to be scanned regularly (daily?) with results being updated across the board.

chigby:
## Description

Fixes #1125

This PR adds some error handling to the scan process in two ways:

1. During the asset extraction phase, if we cannot fetch an asset's contents, skip over that asset and continue. 
2. During a bulk scan, if any error is caught during the scan of a single landing page, skip that landing page and continue.

See commit messages for additional technical details.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing
I've added unit tests for both changes, but it might be useful to try this locally with a bulk scan on the problematic landing page that inspired this issue.

### Post-deployment actions

None.

### General checks

- [x] Verify that the directory scan result page in admin interface loads as expected

### DIFF
--- a/scanner/scanner.py
+++ b/scanner/scanner.py
@@ -92,7 +92,10 @@ def bulk_scan(securedrops: 'DirectoryEntryQuerySet') -> None:
             tldextract.extract(d).registered_domain
             for d in entry.permitted_domains_for_assets
         ]
-        current_result = perform_scan(entry.landing_page_url, permitted_domains)
+        try:
+            current_result = perform_scan(entry.landing_page_url, permitted_domains)
+        except Exception:
+            continue
 
         # This is usually handled by Result.save, but since we're doing a
         # bulk save, we need to do it here
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,5 +1,4 @@
 ---
-version: "3"
 networks:
   app:
 services:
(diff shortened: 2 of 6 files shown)


scanner/scanner.py
```diff
@@ -92,7 +92,10 @@ def bulk_scan(securedrops: 'DirectoryEntryQuerySet') -> None:
             tldextract.extract(d).registered_domain
             for d in entry.permitted_domains_for_assets
         ]
-        current_result = perform_scan(entry.landing_page_url, permitted_domains)
+        try:
+            current_result = perform_scan(entry.landing_page_url, permitted_domains)
+        except Exception:
+            continue
```
SaptakS, line 98: I am slightly worried about this failing too silently. Should we maybe have a warning of some sorts logging the error, in case there is some real error that we may need to debug?

SaptakS (approved):
Looks mostly good to me. Left a small comment, but I don't think that is a blocker for getting this PR merged and deployed. Let's test it on staging.

PR #1153, opened 2024-09-30
Update Landing Page Directory Scan User-Agent to Spoof Firefox on Windows

nathandyer:
Due to some news organizations opting for user agent-level bot preventation tactics, some landing pages were returning a 403 instead of a 200 status code, which triggered a warning in the directory.

To work around this, we can spoof the user agent string for well-supported browsers on popular platforms.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [x] Config changes

## Testing

- [ ] CI passes
- [ ] Verify that the scanner remains functional with the new user agent

### General checks

- [ ] Linting and tests pass locally
- [ ] There is no conflicting migrations


nathandyer:
It appears that the core issue was _not_ with the user agent, but in (likely) blocking of Google Cloud Platform. Closing this PR as not needed.

PR #1170, opened 2025-03-10
Updates Django

SaptakS:
## Description

Updates django to fix vulnerabilities. Though I don't think we are affected by this. https://www.djangoproject.com/weblog/2025/mar/06/security-releases/

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Vulnerabilities update
- [ ] Config changes


chigby (approved):
These updates look good to me!

PR #1199, opened 2025-09-30
Form security improvement

chigby:
Part of https://github.com/freedomofpress/fpf-www-projects/issues/484

## Description
1. Adds the `wagtail-honeypot` package and incorporates it into our forms. It's probably instructive to peruse [the readme documentation](https://github.com/nm-packages/wagtail-honeypot) for it to understand what's going on with it. Essentially: all form pages have a "Honeypot" tab with a checkbox that enables a hidden field. If that hidden field contains any text when submitted, the form response will not be recorded. It's worth testing this out locally as part of code review.
2. The honeypot package includes some JS and CSS, which I've added to the form page template. Notably, the JS manipulates inline style attributes and that needed an update to the content security policy's [`style-src-attr` directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/style-src-attr). If there's a way to get the JS to work without this, I'm not sure what it is, but I feel like it's okay since we're already using some unsafe hashes in there for the wagtail admin.
3. Adds a view that displays a basic error page at the `429/` url.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

1. Make sure you understand how the honeypot package changes Form Pages and test out the form submissions work as desired (in both the "positive" and "negative" cases).
2. Check the 429 page. It looks like this, but maybe you have improvement suggestions?
image

### Post-deployment actions

None.

### General checks

No changes.

- [x] Verify contact form submissions works as expected
- [x] Verify if any CSP changes required (test at least both firefox & chrome)

No changes.
### If it's a major change

Not major.

See above.


mig5 (approved):
Looks good to me @chigby @harrislapiroff ! I have only done a visual review, I'm not in a position at this precise moment to set up a local installation to test the honeypot. But I'll happily test it when it lands in staging.

Great to see the template fix for the 403 route too (which I noticed by accident throws an exception right now in the higher envs, and no doubt was noticed locally too).

Once this lands in the staging/prod environments, I will hook up the 429 logic in the Nginx proxy to load the new page.

mig5:
Got some time to test it locally today, seems great (test both positive and negative form submission).

Since I'm the only reviewer on this, I assume you want me to merge, so I'll do so now.

Then we can adjust the forms in staging to enable the HP setting and test it again there (along with 429 tests that I'll hook up in Nginx)

mig5:
Actually @harrislapiroff or @chigby do you want to handle merge? I note "Merging is blocked - Waiting on code owner review from freedomofpress/fpf-web-developers". I could workaround it but I probably shouldn't? :)

ryanwhitney (approved):
Screenshots lgtm

PR #1240, opened 2026-03-27
Adds middleware to add Onion-Location header with full path

### LINKED ISSUE #1233: Add support for returning the `Onion-Location` header in Wagtail
### Description

To support TOR users on our Wagtail sites after the migration to Traefik, we'll need to add functionality for Wagtail to return the `Onion-Location` header to the client.

The value of the `Onion-Location` header needs to be in this format:
```
{scheme}://{hostname}{path}
```
Where:
* `{scheme}` is `http` (as we don't use SSL connection on our wagtail onion sites)
* `{hostname}` is the onion hostname from `DJANGO_ONION_HOSTNAME`
* `{path}` is the URI of the current resource on the site

For example, the page at `https://securedrop.org/research/` should return this header:
```
Onion-Location: http://sdolvtfhatvsysc6l34d65ymdwxcujausv7k5jk4cy5ttzhjoi6fzvyd.onion/research/
```

### Urgency

High (within the upcoming two-week sprint)

### Impact

Nice-to-Have

### Timing

We are aiming to go-live with the Nginx -> Traefik migration at the end of March (2025-03-31) which is when our existing Nginx controller goes EOL. We could potentially delay a week beyond that but would not want to do anymore than that.

### Notifications

@freedomofpress/fpf-infra

SaptakS:
## Description

Fixes #1233.

Adds a middleware to set Onion-Location header with full path. I have added the middleware only in production settings and not base or dev settings since I am guesseding we are using Onion-Location feature only in prod. But if someone feels strongly that this should always be present (staging and local), I am happy to move it to base.py.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- Run docker compose with the prod-docker-compose.yaml
- Check in network tab of your browser Onion-Location is correctly set along with path.
- I have added some tests as well, ~~but for extensive testing, run ngrok (or some other tunneling, reverse proxy,etc.), open tor browser and check that the Onion-Location button shows up.~~ (The .onion button shows up only if the onion hostname is valid it seems)

### General checks


securedrop/settings/production.py
```diff
@@ -19,7 +19,8 @@
 
 SECRET_KEY = os.environ.get('DJANGO_SECRET_KEY')
 ALLOWED_HOSTS = os.environ.get('DJANGO_ALLOWED_HOSTS').split(' ')
-MIDDLEWARE.insert(1, 'common.middleware.RequestLogMiddleware')  # noqa: F405
+MIDDLEWARE.insert(1, 'common.middleware.request_logger.RequestLogMiddleware')  # noqa: F405
+MIDDLEWARE += ('common.middleware.onion_location.OnionLocationHeaderMiddleware',)  # noqa: F405
```
paulschreiber: question: should this use `MIDDLEWARE.insert()` to match the above, or `MIDDLEWARE.extend()` or `MIDDLEWARE.append()` to match base.py?

Alternately, should this go in `base.py`? `base.py` already has `ONION_HOSTNAME`definition.

common/tests/test_middleware.py
```diff
@@ -104,3 +106,33 @@ def test_request_log_finished(self):
...
+
+    def test_onion_location_correct(self):
+        response = {
+            'status_code': self.response_status_code,
+            'charset': self.response_charset,
+            'reason_phrase': self.response_reason_phrase,
+        }
+        get_response = mock.Mock(return_value=response)
+
+        user_agent = 'Mozilla'
```
paulschreiber: suggestion: this variable is defined but only used once. either inline it or put it in a shared test config.

securedrop/settings/production.py
SaptakS: It can definitely go in `base.py`. Right now I have kept this middleware active only in production deployment and not develop deployment. If we move to `base.py` it would be available in both (but we might not need this in develop).

SaptakS:
Updated the PR addressing your comments.

willbarton (approved):

PR #1255, opened 2026-04-22
fix: update lxml to 6.1.0

paulschreiber:
## Description
Update lxml to 6.1.0. Addresses vulnerability:
[CVE-2026-41066](https://github.com/advisories/GHSA-vfmq-68hx-4jfw)

## Type of change

- [X] Vulnerabilities update


redshiftzero:
This may involving bumping `bs4` too

redshiftzero:
oops accidentally closed

paulschreiber:
Attempting to update to BS 4.14.3 fails
Wagtail 6.3 pins to <4.13:
https://github.com/wagtail/wagtail/blob/stable/6.3.x/setup.py#L31

Wagtail 7 allows BS<5:
https://github.com/wagtail/wagtail/blob/stable/7.3.x/pyproject.toml#L45

paulschreiber:
Upgrading to BS 4.12 does not solve the problem.

willbarton:
*TLDR: The fix was to re-record the VCR.py cassette being used for this test.*

This is a large version jump in lxml, and clearly something about the way it's handling the response content has changed, so I dug in a little bit into what we're doing, what lxml is doing, and why. In [`scanner.request_and_scrape_page`, we pass the response content from a `requests.get` to beautiful soup](https://github.com/freedomofpress/securedrop.org/blob/develop/scanner/scanner.py#L123-L145), and only catch a `requests.exceptions.MissingSchema` error, but lxml is throwing a `ValueError` trying to process namespaces.

We're using VCR.py to store recorded requests, and the request being tested here is a gzipped response. If I use pdb to inspect the failing test I can check the encoding:

```python
-> soup = BeautifulSoup(page.content, "lxml")
(Pdb) page.headers.get("Content-Encoding")
'gzip'
```

I can then unzip the contents and see that it is HTML and I can pass that to `BeautifulSoup` without error. 

```
(Pdb) import gzip
(Pdb) html = gzip.decompress(page.content).decode('utf-8')
(Pdb) soup = BeautifulSoup(html, "lxml")
```

I went down a rabbit hole with lxml that I think is interesting, but that ultimately had nothing to do with the fix here. When we pass the gzipped responses in the VCR.py cassette to lxml/libxml via BeautifulSoup, it doesn't decompress gzipped the strings (which makes sense, not it's job), but tries to process them in weird ways that have changed from version 4.9 to 6.1 and broken this test.

The requests library [automatically decompresses gzip-encoded responses](https://requests.readthedocs.io/en/latest/community/faq/#encoded-data), so the fact that we're getting one here doesn't make sense to handle in our code. It seems to be an artifact from VCR.py. 

I deleted the `scanner/tests/scans_vcr/scan-with-subdomain-redirection.yaml` and let VCR re-record it, and... the compressed HTML is no longer compressed in the recorded file. So, something about how VCR.py handles recording compressed responses must have changed between when this was initially recorded and now. 

But if you're curious ***why this failed now***, well... Until lxml 5.0, it would fail silently when given gzipped contents to process. After that it generated some gobbledygook that wasn't the actual response, and it's not until 6.0 that it generates any errors, but (as we see above) those errors aren't real (when decompressed, the html is successfully processed). 

Note: I only rerecorded the file that's failing. We might want to inspect the other VCR.py cassettes, but that feels out of scope here.

willbarton (approved):

PR #1268, opened 2026-04-29
fix: update Wagtail to 7.4 LTS

### LINKED ISSUE #1248: Upgrade to Wagtail 7.0 LTS

paulschreiber:
# Description
Update Wagtail 6.3.8 to 7.4.0

Resolves #1248

Also update:
* wagtail-factories 4.1.0 > 4.4.0
* wagtail-honeypot 1.2.0 > 1.3.0
* wagtailmedia 0.14.5 > 0.17.2

Part of https://github.com/freedomofpress/fpf-www-projects/issues/508

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [X] Config changes

## Testing

Full QA of site.


paulschreiber:
@willbarton Please both review the code and run this through a regular deploy test/workflow.

blog/models.py
```diff
@@ -29,7 +29,7 @@
 
 class BlogPage(MetadataPageMixin, Page):
     publication_datetime = models.DateTimeField(
-        help_text="Past or future date of publication"
+        null=True, help_text="Past or future date of publication"
```
willbarton, line 32: By setting `null=True` we preserve Wagtail 7.0+'s [deferred validation of pages](https://docs.wagtail.org/en/stable/releases/7.0.html#deferring-validation-on-saving-drafts) when they're in draft. In Wagtail 7.3+ this deferred validation becomes an important part of autosave functionality, so this seems like important user-facing behavior to preserve.

When new pages are created and have never been published, or exists unpublished, the page draft is stored both in a JSON `Revision` as well as directly to the page tables themselves. In order for deferred validation to work, non-text fields need to be null'able in those tables. For published pages, new drafts exist only as JSON `Revision` objects.

Because this `null=True` is not accompanied by a `blank=True` the form validation still requires the field when validation is performed on publish. 

Image

SaptakS (approved):
Looks good to me. Once the branch is rebased, I think this can be merged.

SaptakS:
Let's test this once on staging to verify with real data as well.

PR #1270, opened 2026-05-04
fix: update Python to 3.14

paulschreiber:
## Description
Update Python to 3.14
- updates Wagtail to 7.4 (and wagtail*)
- updates willow to 1.12.0


willbarton (approved):
I think this looks good to me.

One thing I will note, running tests with deprecation warnings on (`python -W always::DeprecationWarning`), we do have a [deprecation warning from Python for `datetime.utcnow`, which is used in our test factories](https://github.com/freedomofpress/securedrop.org/blob/c547d8c06b5263a7da23a9ad1ae5f7f14cf5e46c/directory/tests/factories/entry.py#L97) (we should probably be testing with timezone-aware datetimes).

But we have other deprecation warnings for Django 6 (for URLField https changes) and upstream in dependency packages (predominantly in datetimes and asyncio). I don't think fixing these here should be a blocker.

PR #1271, opened 2026-05-05
fix: no longer pin pip to 25.0.1

paulschreiber:
## Description
No longer pin pip to 25.0.1.

The bug this worked around was fixed a while ago. https://github.com/freedomofpress/securedrop.org/commit/427d660e171ae8a109b6408d6b237ab1bcbcec77


harrislapiroff (approved):
LGTM

PR #1272, opened 2026-05-06
fix: replace psycopg2 with psycopg 3

paulschreiber:
## Description
Replace psycopg2 with psycopg[c]. 
- matches the config on FPF
- psycopg 3 is newer
- psycopg[c] is faster.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes


willbarton (commented):
Is there a reason we want to build the C version each time instead of using `psycopg[binary]`?

paulschreiber:
@willbarton checking with @chigby regarding the C version in https://github.com/freedomofpress/pressfreedomtracker.us/pull/2161

willbarton (approved):

harrislapiroff (approved):
LGTM

PR #1284, opened 2026-05-12
chore: remove manual vulnerability checks

paulschreiber:
## Description
Remove manual vulnerability checks. Avoids duplicate notifications when dependabot is enabled.
- remove safety check script
- remove safety from dependencies file
- update Makefile

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes


harrislapiroff (approved):
Looks like this wasn't running unless invoked manually, so I agree we can get rid of it with no loss of security.

PR #1305, opened 2026-06-01
fix: have django automatically format migrations with ruff

paulschreiber:
## Description
Have django automatically format migrations with ruff using `django-migrations-ruff-formatter`. Avoids manual work by developers.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ lint: ruff
 
 .PHONY: ruff
 ruff: ## Runs ruff linting in Python3 container.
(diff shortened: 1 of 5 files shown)


harrislapiroff (changes requested):
Instead of this, let's add migrations to ruff's ignore list. See conversation here: https://github.com/freedomofpress/pressfreedomtracker.us/pull/2219#pullrequestreview-4453800796

PR #1318, opened 2026-06-09
docs: update copyright date on LICENSE

paulschreiber:
## Description
update copyright date on LICENSE

update LICENSE text to match other repos.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -3,178 +3,178 @@ Copyright © 2017–2026 Freedom of the Press Foundation and contributors
                     GNU AFFERO GENERAL PUBLIC LICENSE
                        Version 3, 19 November 2007
 
-Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (C) 2017- Freedom of the Press Foundation and contributors
```
harrislapiroff, line 1: I actually think this is better open-ended? It implies "through present" and we don't have to update it every year. (This also matches the securedrop has in their license https://github.com/freedomofpress/securedrop/blob/develop/LICENSE)

harrislapiroff (changes requested):

LICENSE
paulschreiber: I don't believe an open-ended range is valid.

paulschreiber: Trevor confirmed we should use a single year or a range of years.

harrislapiroff (approved):

PR #1337, opened 2026-06-16
fix: prevent dependabot from updating to a major django or wagtail version

paulschreiber:
## Description
Prevent dependabot from updating to a major Django version. Allow minor and patch versions.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -33,3 +33,8 @@ updates:
       - dependency-name: 'django'
         update-types:
           - 'version-update:semver-major'
+          - 'version-update:semver-minor'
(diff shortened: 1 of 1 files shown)


SaptakS (changes requested):
Given Wagtail and Django have LTS in minor versions, I feel like we should only allow patch version updates else we might move out of LTS I think? This does mean we would still need to do LTS updates manually which I think is reasonable.

SaptakS (approved):
Looks good to me.

PR #1354, opened 2026-06-21
Fix utc in securedrop.org

zrobert7:
## Description

Fixes https://github.com/freedomofpress/fpf-www-projects/issues/546 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Write out any special testing steps here.

### Post-deployment actions

In case this PR needs any admin changes or run a management command after deployment, mention it here:

### General checks

- [ ] Linting and tests pass locally
- [ ] There is no conflicting migrations

### If it's a major change

If the PR involves some visual changes in the frontend, it is recommended to add a screenshot of the new visual.

### DIFF
--- a/directory/tests/factories/entry.py
+++ b/directory/tests/factories/entry.py
@@ -1,31 +1,34 @@
 import random
 import string
-from datetime import datetime, timezone
 
 import factory
 import wagtail_factories
-
+from django.utils import timezone
 from faker import Faker
+
 from common.models import CustomImage
 from directory.models import DirectoryEntry, ScanResult
 from directory.tests.factories.taxonomy import (
-    LanguageFactory,
     CountryFactory,
+    LanguageFactory,
     TopicFactory,
 )
 
 
 def random_onion_address():
-    return 'https://' + ''.join(
-        random.choice(string.ascii_lowercase + string.digits)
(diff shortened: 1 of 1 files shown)


directory/tests/factories/entry.py
```diff
@@ -38,63 +35,63 @@ class Meta:
...
-            self._prefetched_objects_cache = {"topics": topics}
+            self._prefetched_objects_cache = {'topics': topics}
 
 
 class ScanResultFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = ScanResult
 
-    result_last_seen = factory.LazyFunction(datetime.utcnow)
+    result_last_seen = factory.LazyFunction(lambda: datetime.now(timezone.utc))
```
paulschreiber: Can you use this? Also eliminates the change to import.
```suggestion
    result_last_seen = factory.LazyFunction(datetime.timezone.utc)
```

willbarton: I don't think we need to use a lambda here, we can use Django's `timezone.now()` which is UTC-aware.

willbarton (approved):

PR #1371, opened 2026-06-30
fix: replace pip with poetry

paulschreiber:
## Description
Use poetry as a python package manager instead of pip.
- move dependencies to pyporject.toml
- remove requirements files
- add poetry.lock
- update Makefile, README and CLAUDE documentation

Works alongside https://github.com/freedomofpress/actionslib/pull/56

@enpaul @mig5 Please look over the Dockerfile changes. Gemini helped write this part. Having a human review (and adjust as needed) would be much appreciated.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes


devops/docker/DevDjangoDockerfile
```diff
@@ -3,24 +3,26 @@ FROM python:3.14.6-slim-trixie@sha256:44dd04494ee8f3b538294360e7c4b3acb87c8268e4
...
-COPY --from=builder dev-requirements.txt ./
-COPY --from=builder requirements.txt ./
+ENV POETRY_NO_INTERACTION=1 \
+    POETRY_VIRTUALENVS_IN_PROJECT=1 \
+    POETRY_VIRTUALENVS_CREATE=1
+
+# MATCHING WORKDIR: Keep this the same as the final stage
+WORKDIR /django
 
+RUN pip install --no-cache-dir poetry
```
mig5: There's no signature validation of installing poetry itself here - which everything else then relies on as having been not compromised. (The classic 'catch 22' of 'who validates the validator'.)

Since you are building from a Trixie image, you might be able to apt-get install `python3-poetry` within lines 4-8 (assuming that its Poetry 2.1 will be sufficient for the needs here - I *think* it will?)

devops/docker/ProdDjangoDockerfile
```diff
@@ -1,3 +1,4 @@
+# Stage 1: Build node assets (Unchanged)
```
mig5: Just as a nit I would remove the '(Unchanged)' bit here, that's AI trying to reassure.

You could optionally consider adding `# Stage 1: Build node assets` to the DevDjangoDockerFile for consistency, though.

devops/docker/ProdDjangoDockerfile
```diff
@@ -15,9 +16,31 @@ RUN cd /src-files && ( npm install && npm run build )
 LABEL MAINTAINER="Freedom of the Press Foundation"
 LABEL APP="securedrop.org"
 
+# Stage 2: Poetry Builder (Renamed alias to python-builder)
```
mig5: Again (it's a nitpick) I'd just remove the parenthesis section as it's trying to tell you what it changed - won't make sense post-merge.

devops/docker/ProdDjangoDockerfile
```diff
@@ -15,9 +16,31 @@ RUN cd /src-files && ( npm install && npm run build )
...
+    libxml2-dev \
+    libxslt-dev \
+    libz-dev \
+    python3-dev \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+ENV POETRY_NO_INTERACTION=1 \
+    POETRY_VIRTUALENVS_IN_PROJECT=1 \
+    POETRY_VIRTUALENVS_CREATE=1
+WORKDIR /django
+RUN pip install --no-cache-dir poetry
```
mig5: Ditto per what I wrote for `DevDjangoDockerFile` - if a compromised poetry is installed by pip, you can't necessarily trust what it said about the poetry.lock SHAs being correct either. Hopefully we can apt install Poetry?

devops/docker/ProdDjangoDockerfile
```diff
@@ -47,12 +69,18 @@ RUN find /django -path /django/node_modules -prune -o -print -exec chown gcorn:
 
 WORKDIR /django
 ENV TLDEXTRACT_CACHE="/tmp/python-tldextract"
-RUN pip install --no-deps --require-hashes -r /django/requirements.txt
 
+# --- POETRY INTEGRATION ---
+# FIXED: Copying from python-builder instead of python-base
```
mig5: Ditto.. just drop the 'FIXED' bit

devops/scripts/version-file.sh
```diff
@@ -42,15 +51,15 @@ ${branch_desc}
...
+${git_history}
 
 #### PYTHON INFO ####
 
 $(python3 --version)
 
 #### PYTHON DEPS ####
 
-$(pip freeze)
+$(poetry show --no-ansi --no-truncate)
```
mig5: This step fails in ProdDjangoDockerFile because there is no `poetry` binary by the time it gets to this stage (it only got installed in the `python-builder` stage):

```
#34 [python-base 17/18] RUN /django/devops/scripts/version-file.sh
#34 0.124 /django/devops/scripts/version-file.sh: line 47: poetry: command not found
#34 DONE 0.1s
```

I know the point is to get rid of `pip`, but for the sake of a simple 'display' of dependencies (not relying on pip itself for package *management*, the simplest solution might be to just still use pip for this one call here, like this:

```
$(/django/.venv/bin/python -m pip freeze)
```

The output to `/deploy/version-full.txt` will still look the way you want it.

pyproject.toml
```diff
@@ -1,6 +1,48 @@
 [project]
 name = "securedrop.org"
+version = "0.1.0"
+package-mode = false
```
mig5: In modern Poetry, the [docs](https://python-poetry.org/docs/basic-usage/#operating-modes) say that `package-mode` should go in the `[tool.poetry]` section. I think that's true of Poetry 2 and up.

```
[tool.poetry]
package-mode = false
```

It doesn't error, though, so maybe it's forgiving.

docker-compose.yaml
```diff
@@ -59,6 +59,8 @@ services:
     userns_mode: keep-id
     volumes:
       - ./:/django:z
+      # FIX: Protects and mounts the image's built .venv so host files don't override it
```
mig5: Again me and my nitpicking over AI's verbosity habits: I'd drop the `FIX`.

mig5 (changes requested):
Nice one @paulschreiber . A few things I spotted, mostly nitpicks, plus one or two real issues, but overall looks good, hopefully straightforward to sort out the poetry install and the fix for the deps writing to the txt file.

devops/docker/ProdDjangoDockerfile
paulschreiber:
```suggestion
# Stage 1: Build node assets
```

devops/docker/ProdDjangoDockerfile
paulschreiber:
```suggestion
# Stage 2: Poetry Builder
```

devops/docker/ProdDjangoDockerfile
paulschreiber:
```suggestion
```

docker-compose.yaml
paulschreiber:
```suggestion
```

mig5:
@paulschreiber This should be good now. 

Per our discussions, the most important thing is to establish 'trust provenance' for the poetry install, otherwise nothing else poetry does can be trusted.

The most ideal scenario would be apt install python3-poetry, however it unfortunately lacks patches for certain vulnerabilities which we think matter.

To answer your other question, `pipx` also doesn't help here (it doesn't do anything different to `pip` in terms of cryptographic attestation of the signatures of the package you're installing), it mostly just manages separate virtual environments per installed package (solving a different sort of challenge in package management).

The Infra team solves this in their own repo by:

a. Making poetry core itself a dev dependency in pyproject.toml
b. Generating a 'requirements' file *from* the poetry.lock file using the `poetry export` command
c. Using that requirements file as a one-time pip install command to install poetry itself
d. Then use poetry as normal for installing stuff from poetry.lock.

I've adopted that same process for this repo. Here are the main changes I made:

1. We generate a new file `bootstrap-poetry.txt` (essentially a 'requirements.txt' style file) from the poetry.lock file via the 'poetry export' plugin (I also added this as a dev dependency to pyproject.toml). This is a requirements file to get just what it takes to install poetry itself, via pip. It is always worth double-checking at Pypi that the hashes look okay (not perfect as a check, if the Pypi version itself is compromised, but better than nothing).
2. Added a pre-commit step that routinely regenerates the `bootstrap-poetry.txt` when `poetry.lock` itself changes. It's not 100% clear to me if Dependabot will pick that up (as it probably isn't running `pre-commit`), but you might be able to add a rule that handles tracking dependencies in `bootstrap-poetry.txt` separately. At the very least, the `pre-commit` step will be useful for when it's an actual developer who has run a `poetry update` or `poetry lock` command.
3. Adapted the Dockerfiles to copy the `bootstrap-poetry.txt` file into the staged build and then use it to install poetry
4. While I was there, fixed an observation I had yesterday whereby the `version-file.sh` script didn't fail the Docker build when it itself failed (which could otherwise go unnoticed if something goes wrong there at a later date).

I tested building both images and it succeeds:

```
docker compose --progress=plain build
docker compose -f prod-docker-compose.yaml --progress=plain build
```

Recommend you test yourself too to confirm.

/cc @enpaul FYI

paulschreiber:
@mig5 @enpaul I don't believe dependabot will keep `bootstrap-poetry.txt` up to date automatically. It looks like renaming it to `bootstrap-poetry-requirements.txt` will work. Is that okay?

mig5:
@paulschreiber If that works for Dependabot, then yep, absolutely! I didn't realise it matched names in that way. The naming of the file is purely cosmetic otherwise here from a workflow perspective :+1:

devops/docker/DevDjangoDockerfile
```diff
@@ -36,20 +46,28 @@ RUN apt-get update && apt-get install -y \
...
-COPY --from=builder dev-requirements.txt /requirements.txt
+RUN chmod +x /usr/local/bin/django-start.sh
 
+WORKDIR /django
 ENV TLDEXTRACT_CACHE="/tmp/python-tldextract"
-RUN pip install --no-deps --require-hashes -r /requirements.txt
 
+# --- POETRY INTEGRATION ---
+# MATCHING PATH COPY: Copying /django/.venv to /django/.venv preserves shebang paths
+COPY --from=builder /django/.venv /django/.venv
```
enpaul: Copying virtualenvs is a really bad idea as it creates inherrent dependencies between the source and destination filesystems, shared libraries, and python installations. From the [`venv` docs](https://docs.python.org/3/library/venv.html):

> Warning
>
> Because scripts installed in environments should not expect the environment to be activated, their shebang lines contain the absolute paths to their environment’s interpreters. Because of this, environments are inherently non-portable, in the general case.

Further, copying the entire venv directory tree will also definitionally copy any sdists that were downloaded and built during package installation which will unnecessarily inflate the final size of the image. Instead of copying the venv, you should recreate it in the final stage and install the built packages to it. This ensures that venv is always correct for the filesystem that it exists on, any shared libraries are linked correctly, and the venv tree includes only the wheel package data and no build artifacts.

devops/docker/ProdDjangoDockerfile
```diff
@@ -15,9 +16,40 @@ RUN cd /src-files && ( npm install && npm run build )
...
+    --require-hashes \
+    -r bootstrap-poetry-requirements.txt && \
+  ln -sf ~/.local/share/poetry/bin/poetry /usr/local/bin/poetry
 
+COPY pyproject.toml poetry.lock ./
+# Poetry natively validates cryptographic hashes from poetry.lock
+RUN poetry install --without dev --no-root
+
+# Stage 3: Final Production Image
+FROM python:3.14.6-slim-trixie@sha256:44dd04494ee8f3b538294360e7c4b3acb87c8268e4d0a4828a6500b1eff50061 AS python-base
```
enpaul: I'm not sure I understand why this stage is labeled "python-base" since it's not used as a base image in any other stage?

devops/docker/ProdDjangoDockerfile
```diff
@@ -47,12 +78,17 @@ RUN find /django -path /django/node_modules -prune -o -print -exec chown gcorn:
 
 WORKDIR /django
 ENV TLDEXTRACT_CACHE="/tmp/python-tldextract"
-RUN pip install --no-deps --require-hashes -r /django/requirements.txt
 
+# --- POETRY INTEGRATION ---
+COPY --from=python-builder /django/.venv /django/.venv
```
enpaul: Same note from the dev image about copying the venv. This may work right now, but if the underlying image changes, the available sdists/wheels for dependent packages change, or this path is changed in the future this will break in subtle and hard to diagnose ways.

devops/docker/ProdDjangoDockerfile
```diff
@@ -72,9 +108,17 @@ USER ${USERID}
...
-# Run version info script as gcorn user (required as the repo directories are owned by this user)
+# Inject host Git info as build args so version-file.sh can use them as fallbacks
+ARG GIT_COMMIT_SHA
+ARG GIT_BRANCH
+ARG GIT_TAG
+ENV GIT_COMMIT_SHA=${GIT_COMMIT_SHA} \
+    GIT_BRANCH=${GIT_BRANCH} \
+    GIT_TAG=${GIT_TAG}
+
 RUN /django/devops/scripts/version-file.sh
```
enpaul: I'm pretty sure these directives can be combined like below:


```suggestion
RUN GIT_COMMIT=${GIT_COMMIT_SHA} GIT_BRANCH=${GIT_BRANCH} GIT_TAG=${GIT_TAG} /django/devops/scripts/version-file.sh
```

devops/docker/ProdDjangoDockerfile
```diff
@@ -15,9 +16,40 @@ RUN cd /src-files && ( npm install && npm run build )
...
+RUN python3 -m venv ~/.local/share/poetry && \
+  ~/.local/share/poetry/bin/python -m pip install \
+    --only-binary=:all: \
+    --require-hashes \
+    -r bootstrap-poetry-requirements.txt && \
+  ln -sf ~/.local/share/poetry/bin/poetry /usr/local/bin/poetry
 
+COPY pyproject.toml poetry.lock ./
+# Poetry natively validates cryptographic hashes from poetry.lock
+RUN poetry install --without dev --no-root
```
enpaul: Rather than explicitly excluding the `dev` group, I'd recommend using `--only=main` to explicitly only include the primary package dependencies. This ensures that if future dependency groups are added their dependencies won't be installed unless explicitly included here.

devops/scripts/version-file.sh
```diff
@@ -1,13 +1,11 @@
 #!/bin/bash
```
enpaul, line 1: An important behavior change in this script is that it will no longer exit with an error (causing a failed build) if the git info cannot be determined. The script will instead populate the git debug info with various "unknown" strings. Is that ok, or should it force a build failure if the git info cannot be determined?

devops/scripts/version-file.sh
```diff
@@ -16,22 +14,38 @@ trap 'handle_error $LINENO $?' ERR
...
+    commit="${GIT_COMMIT_SHA:-unknown}"
+    ref_desc="tag/release: ${GIT_TAG:-unknown}"
+    git_history="Git history unavailable in this build environment."
+fi
+
+# Trim commit to short hash if provided via environment variable
+commit="${commit:0:7}"
+
+python_version="$(python3 --version)"
+python_deps="$(/django/.venv/bin/python -m pip freeze)"
```
enpaul: I think the python debug info and package debug info should be sourced from the same interpreter to avoid potential issues that could arise from mismatches in the future.

enpaul: I'd also recommend not hardcoding this, or at least parameterizing it so that if the venv path is ever changed or a non-venv approach is used in the future it doesn't break this script as well.

docker-compose.yaml
```diff
@@ -59,6 +59,7 @@ services:
     userns_mode: keep-id
     volumes:
       - ./:/django:z
+      - /django/.venv
```
enpaul: What's the purpose of this volume mount?

pyproject.toml
```diff
@@ -1,6 +1,51 @@
 [project]
 name = "securedrop.org"
+version = "0.1.0"
```
enpaul, line 3: Will the package version be used moving forward? That will be really cool to be able to correlate a package version with a wheel or sdist artifact

enpaul (changes requested):
Thanks for putting this together @paulschreiber! I've left some requests for changes and also asked some questions below

devops/scripts/version-file.sh
paulschreiber: Which one of these do you prefer? @mig5 @enpaul freel free to push a change here with your preferred method.

devops/docker/ProdDjangoDockerfile
paulschreiber: What do you recommend calling it?

pyproject.toml
paulschreiber: We have not versioned websites in the past. If this is something that would be useful, I'd be interested in learning more.

docker-compose.yaml
paulschreiber: From Gemini:
>When you use a bind mount like .:/django, Docker overlays your host machine's folder directly on top of the container's /django folder at runtime.
>
>Because Poetry was configured with POETRY_VIRTUALENVS_IN_PROJECT=1, it built the virtual environment inside the image at /django/.venv. When Docker Compose mounts your local directory over /django, it hides the .venv folder that was built inside the image and replaces it with whatever is on your host machine (which either doesn't have a .venv folder, or has a mismatched host-OS .venv).
>
>You need to tell Docker Compose to preserve the container's internal .venv folder by giving it its own anonymous volume. This prevents your host directory from overwriting it.

If there is a better way to handle this, please change it.

devops/docker/DevDjangoDockerfile
paulschreiber: What's the right way to do this? Can you update the PR with a better method?

devops/docker/ProdDjangoDockerfile
paulschreiber: Please push the change here as well.

pyproject.toml
willbarton: I wonder if this is a chance to use https://github.com/mtkennerly/poetry-dynamic-versioning, since we already create release tags — but it might be out-of-scope for this initial work since I'm not sure how right off it plays with our date-based tagging instead of semver (semver doesn't make sense to me for a website).

devops/docker/ProdDjangoDockerfile
chigby: It's possible the name python-base was added in a previous incarnation of this file. But if it's not being used, the `AS <name>` part can be removed, as it's entirely optional.

mig5:
@enpaul Please see three commits I made:

____

https://github.com/freedomofpress/securedrop.org/pull/1371/commits/88deffc3ace4b9023cf768527deb448543adf7fc

This hopefully solves Ethan's concerns about the venvs:

> Further, copying the entire venv directory tree will also definitionally copy any sdists that were downloaded and built during package installation which will unnecessarily inflate the final size of the image. Instead of copying the venv, you should recreate it in the final stage and install the built packages to it. This ensures that venv is always correct for the filesystem that it exists on, any shared libraries are linked correctly, and the venv tree includes only the wheel package data and no build artifacts.

Indeed what should happen now is we build wheels during the builder stage and then install from those.

It also removes the mounting of the `.venv` into the container as that shouldn't be required anymore. The virtualenv that gets copied in goes into `/opt/venv` and also has the bonus of not being chowned to gcorn, so the runtime user can't manipulate it.

____

https://github.com/freedomofpress/securedrop.org/pull/1371/commits/e77a509e435ebe76848333192591851f8990c7d6

Reinstated the mandatory git commit sha/branch/tag for the prod build

_____

https://github.com/freedomofpress/securedrop.org/pull/1371/commits/15fa7cf4bce352f58161ba624a2e184248cba0d9

Relaxing that last change for GIT_TAG because otherwise CI fails... hopefully will be ok for actual prod builds and still send the appropriate tag? Not 100% sure.

paulschreiber:
Dependencies have been updated substantially (additional, removals, updates) since this was created. We'll need to re-import the contents of requirements and dev-requirements into pyproject.toml.

devops/docker/ProdDjangoDockerfile
```diff
@@ -10,9 +10,55 @@ RUN cd /src-files && ( npm install && npm run build )
...
+    python3-dev \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+ENV POETRY_NO_INTERACTION=1
+
+WORKDIR /django
 
+COPY bootstrap-poetry-requirements.txt ./
+RUN python3 -m venv ~/.local/share/poetry && \
+  ~/.local/share/poetry/bin/python -m pip install \
+    --only-binary=:all: \
```
enpaul: In most practical terms this shouldn't cause any problems, but this will technically cause the build to fail if one of the poetry dependencies doesn't have a wheel available on pypi. Might be worth relaxing this by using `--prefer-binary=:all:`.

devops/docker/ProdDjangoDockerfile
```diff
@@ -10,9 +10,55 @@ RUN cd /src-files && ( npm install && npm run build )
...
+      --no-deps \
+      --wheel-dir=/wheels \
+      -r /tmp/requirements.txt && \
+    find /wheels -maxdepth 1 -type f -name '*.whl' -print | \
+      LC_ALL=C sort | \
+      while IFS= read -r wheel; do \
+        printf '%s --hash=sha256:%s\n' \
+          "$wheel" \
+          "$(sha256sum "$wheel" | cut -d' ' -f1)"; \
+      done > /tmp/wheel-requirements.txt
```
enpaul, line 58: This is really clever, I've never seen this approach of dynamically creating a req.txt from directory contents before, I like this.

devops/docker/ProdDjangoDockerfile
```diff
@@ -42,13 +87,28 @@ COPY --from=node-assets /src-files/ /django/
...
-
+ENV TLDEXTRACT_CACHE="/tmp/python-tldextract" \
+    VIRTUAL_ENV="/opt/venv" \
+    PATH="/opt/venv/bin:$PATH"
+
+# Create the environment in the filesystem where it will run. Installing from
+# the wheelhouse avoids copying a non-portable virtualenv or build artifacts.
+COPY --from=dependency-builder /wheels /wheels
+COPY --from=dependency-builder /tmp/wheel-requirements.txt /tmp/wheel-requirements.txt
+RUN python3 -m venv "${VIRTUAL_ENV}" && \
```
enpaul, line 98: Possibly a controversial take, but I don't think there's actually any benefit to using a python venv in a container. In a single app environment (like a container) dependency isolation is a non-issue, and the installation to the system site-packages via `root` already isolates ownership/permissions from the runtime user so there's no security benefit either. Meanwhile, it takes up (an admittedly small amount of) extra space in the image and adds (an also admittedly small amount of) cognitive overhead to remember/discover where the venv/site-packages is when debugging.

This is also maybe something that could be changed as part of the larger containerization rewrite. It's very much not hurting anything, but wanted to call it out.

enpaul (approved):
Looks good to me! Thanks @mig5 for making those changes and @paulschreiber for putting this together. Few nitpicky things highlighted below, nothing blocking but just stuff I wanted to highlight

devops/docker/ProdDjangoDockerfile
mig5: I think in the interests of getting this old PR finally merged in (which is now starting to cause other problems such as dependabot targeting the `requirements.txt` in the `develop` branch, which this PR removes), I think we should just accept how it is right now.

I know there is a big docker container build overhaul happening by @conorsch , perhaps the venv matter can be addressed more broadly across the Wagtail site builds then.

devops/docker/ProdDjangoDockerfile
mig5: > Might be worth relaxing this by using --prefer-binary=:all:.

Apparently `--prefer-binary` doesn't take a value, but I'll adjust accordingly, thanks.

mig5:
OK, an update:

1. I've applied Ethan's requested change for `--preferred-binary` instead of `--only-binary`, 
2. I've brought the `poetry.lock` up to date (I think) with the latest dependencies that have been getting merged into `develop` branch by way of Dependabot changing `requirements.txt` (that file is being deleted in this Poetry PR - we want the equivalent changes in the `poetry.lock` here). **NOTE**: I may still have missed some, but it's too awkward: *we are better off merging this PR and then merging any follow-up PRs that dependabot makes, which target `poetry.lock`, IMO.*
3. I got the `bootstrap-poetry-requirements.txt` updated accordingly (it is to be re-generated via pre-commit when `poetry.lock` changes, just like the infrastructure repo was doing before we moved to nix).
4. Added a GHA workflow that automatically updates the `bootstrap-poetry-requirements.txt` as a fresh PR when `poetry.lock` changes would result in the need to change the requirements file. The scenario I'm trying to cover here is when Dependabot updates the `poetry.lock`, but isn't able to use our `.pre-commit-config.yaml` to auto-update the requirements file via a precommit hook like a developer workstation would.

Ethan's venv concerns aside, I think at this point that we should just merge this in so that we can focus on the bigger overhauls of container image building, and also so that Dependabot starts targeting the `poetry.lock` moving forward.

@conorsch has been briefed and will hopefully take a look at this early this week.

.github/workflows/sync-poetry-bootstrap.yaml
```diff
@@ -0,0 +1,75 @@
...
+        run: |
+          python -m venv "${RUNNER_TEMP}/poetry-bootstrap"
+
+          "${RUNNER_TEMP}/poetry-bootstrap/bin/python" -m pip install \
+            --require-hashes \
+            --only-binary=:all: \
+            --no-deps \
+            --requirement bootstrap-poetry-requirements.txt
+
+          echo "${RUNNER_TEMP}/poetry-bootstrap/bin" >> "${GITHUB_PATH}"
```
conorsch, line 47: The CI workflow to _automatically_ bump lockfiles for poetry is quite slick, but I'd prefer to avoid having to maintain a separate lockfile just for poetry, if we can get away with it. I realize a lot of discussion in this PR has centered on how best to ensure we're pinning predictable versions everywhere, and I agree with that focus, have Ideas™ on how to achieve it. Notably, the only other place we were using the "bootstrap-poetry-requirements.txt"—the private infra repo—recently dropped this convention. So I'd prefer to land on a different approach, unless we're absolutely forced to resurrect this pattern.

devops/docker/DevDjangoDockerfile
```diff
@@ -1,25 +1,47 @@
...
+    -r bootstrap-poetry-requirements.txt && \
+  ln -sf ~/.local/share/poetry/bin/poetry /usr/local/bin/poetry
+
+COPY pyproject.toml poetry.lock ./
+# Match `poetry install --no-root`: include all non-optional dependency groups.
+# BuildKit transfers only wheels and the hashed export to the final stage.
+RUN poetry export \
+      --format=requirements.txt \
+      --output=/tmp/requirements.txt \
+      --all-groups && \
```
conorsch, line 29: I'm not convinced we need poetry in the container at all, which is a significant portion of the diff of this PR. On L26, we invoke `poetry` to generate a lockfile of requirements—which should have been version-controlled directly in the repo as a first-class part of the build context—then later on L71 we run `pip install -r <requirements>`. So we didn't really gain any simplification by pulling in poetry, and we're still transforming declarations of requirements—which I would argue is actually the problem we should target for simplification: _reducing_ the machinery used to reason about versions and their hashes.

devops/scripts/version-file.sh
```diff
@@ -16,22 +14,55 @@ trap 'handle_error $LINENO $?' ERR
...
-        branch_desc="no branch, detached HEAD"
-        ref_desc="not tagged, $(git describe)";;
-    *)
-        branch_desc="on branch: ${branch}"
-        ref_desc="not tagged, $(git describe)";;
-esac
+# Check if we are inside a valid git repository
+if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
+    branch="$(git rev-parse --abbrev-ref HEAD)"
+    commit="$(git rev-parse --verify HEAD)"
```
conorsch, line 20: Love the nod to git worktrees. I use them regularly, and not having first-class support for them in the version-file script has bitten me in the past, too.

pyproject.toml
```diff
@@ -1,6 +1,49 @@
...
+license = "AGPL-3.0-or-later"
 requires-python = ">=3.14"
+dependencies = [
+    "cryptography (>=49.0.0)",
```
conorsch, line 7: Tracking the deps in `pyproject.toml` is a big win from a devex standpoint: it's way easier to reason about what's installed where.

conorsch (changes requested):
I know I'm late to the review party; forgive me for chiming in after so much discussion has already been hashed out. 

This PR touches a lot of the containerfile overhaul logic that I've started over in https://github.com/freedomofpress/freedom.press/pull/3173. In terms of work sequencing, I'd very much like to land the containerfile reorg first, then target a much more minimal diff on the python package management stream.

So, even though this PR has been through a lot, I'm basically asking for yet more time: let's pare down the scope of this work, land the containerfile changes separately, then revisit just the python deps changes in isolation.

paulschreiber:
Closing this. Will get rewritten after the container refactor/cleanup.

PR #1397, opened 2026-07-02
fix: serve favicon.ico and robots.txt from WHITENOISE_ROOT 

paulschreiber:
## Description
Serve `favicon.ico` and `robots.txt` from `WHITENOISE_ROOT `.

Remove old template view for robots.txt

Partially addresses https://github.com/freedomofpress/fpf-www-projects/issues/581

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes


securedrop/settings/base.py
```diff
@@ -106,6 +106,7 @@
     STORAGES["staticfiles"] = {
         "BACKEND": "whitenoise.storage.CompressedManifestStaticFilesStorage",
     }
+    WHITENOISE_ROOT = os.path.join(BASE_DIR, "securedrop", "static")
```
harrislapiroff: Ack, I hate to be so nitpicky about this multiple times, but I think this isn't the right solution either for two reasons:

1. The result of this is that our javascript files in that directory will now *also* be served at the root (so `piwik.js` will be available at `https://securedrop.org/static/js/piwik.js` *and* `https://securedrop.org/js/piwik.js` and, while I can't think offhand of any problems that would cause, it's at least weird. (It would cause a problem if someone added a folder name that collided with a page URL, e.g., if someone added static files relating to the directory to `/securedrop/static/directory/` that would cause the directory at https://securedrop.org/directory/ to stop working in a way that might confuse a future dev. This is a bit contrived, since that probably isn't where they would add such files.)
2. More importantly I think this directory structure kinda muddies the distinction between the files we *want* to serve at the root and the ones we *want* to serve under `/static/` which could be potentially confusing to a future developer more distant from this change.

I'd suggest either:

1. Move favicon and robots to `/securedrop/static/root/` and make this `WHITENOISE_ROOT = os.path.join(BASE_DIR, "securedrop", "static", "root")`. The root files will still end up getting served twice, but at least the intention will be clear—and polluting the `/static/` namespace is less high-stakes than polluting the root of the site.
2. Create a separate directory at `/root_static/` or similar and put the files in there and change this to `WHITENOISE_ROOT = os.path.join(BASE_DIR, "root_static")`. I prefer this solution for minimal ambiguity, but I'm not wedded to it.

harrislapiroff (dismissed):

zrobert7 (approved):

harrislapiroff (approved):

PR #1498, opened 2026-07-17
ci: don't pull an image just for curl

paulschreiber:
## Description
Same as https://github.com/freedomofpress/freedom.press/pull/3089, but for SDO
We were pulling an external image simply to run `curl` against the django service. We already have curl inside the django container, so let's just use it.

<details>
<summary>
example CI failure logs
</summary>

```
Run docker run --pull=missing --rm --network=freedompress_app "docker.io/curlimages/curl:latest" \
  docker run --pull=missing --rm --network=freedompress_app "docker.io/curlimages/curl:latest" \
  --ipv4 --retry 24 --retry-delay 5 --retry-all-errors http://django:8000/health/ok/
  shell: /usr/bin/bash -e {0}
  env:
    VERSION_DOCKER: latest
    COMPOSE_FILE_DEV: docker-compose.yaml
    COMPOSE_FILE_PROD: prod-docker-compose.yaml
Unable to find image 'curlimages/curl:latest' locally
latest: Pulling from curlimages/curl
8015b0f4edbc: Pulling fs layer
8015b0f4edbc: Download complete
8015b0f4edbc: Pull complete
Digest: sha256:7c12af72ceb38b7432ab85e1a265cff6ae58e06f95539d539b654f2cfa64bb13
Status: Downloaded newer image for curlimages/curl:latest
curl: (6) Could not resolve host: django
Warning: Problem : timeout. Retrying in 5 seconds. 24 retries left.
```
</details>

A few things in [that failed run](https://github.com/freedomofpress/freedom.press/actions/runs/29507956189/job/87658450175) stand out immediately:

1. We're pulling a container image from a remote registry in order to run `curl`, when our images already have it.
2. We're communicating cross-network rather than intra-network to poll the healthcheck endpoint.
3. We're manually running a bespoke command to poll for app readiness, rather than using a first-class healthcheck block via compose.

The solution in this PR addresses all of those points, and should prove more stable on future CI runs.

## Scope of changes

This PR only affects CI behavior, in order to reduce failure rate, so no concerns for live deployments.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -109,16 +109,7 @@ jobs:
         run: make dev-init
 
       - name: Start
-        run: |
-          docker compose --file="$COMPOSE_FILE_DEV" up --wait --pull=missing --no-build
-
-          # Check for failed services and error if any are found
-          sleep 5
-          if [[ $(docker compose --file="$COMPOSE_FILE_DEV" ps --status='[restarting, dead, exited]' --format=json) ]]; then
-            echo "ERROR: One or more services failed to start"
-            docker compose ps
-            exit 1
-          fi
+        run: docker compose --file="$COMPOSE_FILE_DEV" up --wait --wait-timeout=180 --detach --pull=missing --no-build
 
       - name: Make target ${{ matrix.target }}
         run: make ${{ matrix.target }}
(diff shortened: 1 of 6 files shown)


chigby (approved):
Everything here looks good from my point of view.

PR #1507, opened 2026-07-20
fix: add ruff as a dev requirement and simplify Makefile

paulschreiber:
## Description
Add ruff as a dev requirement and simplify Makefile. Matches behavior of FPF/PFT.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes


harrislapiroff (approved):
lgtm

PR #1566, opened 2026-08-05
build: overhaul container build logic

conorsch:
## Description

This PR attempts to standardize the container build logic for this repo, matching the pattern already approved and merged in https://github.com/freedomofpress/freedom.press/pull/3173. Specifically, the changes are:

* use a single containerfile with a multi-stage build, copying assets between
* organizes deployment build tooling under top-level `ci/` directory
* rewrites the `Makefile` as a `justfile` :sparkles: 
* various CI touch-ups

What this PR does _not_ include:

* pip -> poetry conversion, a la https://github.com/freedomofpress/securedrop.org/pull/1371
* comprehensive dev env tooling (will be easier with switch from pip -> poetry)
* clean separate of unit and integration testing (best handled in a separate effort)

I'm intentionally punting on those features until we land the container changes in the Wagtail repos, then I'll do another narrowly-scoped pass.

## Related work
* ports https://github.com/freedomofpress/freedom.press/pull/3173
* refs https://github.com/freedomofpress/freedom.press/pull/3235
* towards https://github.com/freedomofpress/infrastructure/issues/6967

## Testing and review

Please make sure the local compose setup works for you! Can you run the operations you usually do? Is anything broken or surprising?

For the most part, the commits are discrete enough to snip out and drop if anything's objectionable.


paulschreiber:
I'm also updating FPF on fix/containers to make the files match more closely to what you're doing here.

conorsch:
Ready for review! I've rebased on top of latest develop, in particular to include https://github.com/freedomofpress/securedrop.org/pull/1556, so make sure to repull before appending anything.

.github/workflows/publish.yaml
```diff
@@ -22,7 +22,8 @@ jobs:
       packages: write
     with:
       context: '.'
-      containerfile: devops/docker/ProdDjangoDockerfile
+      containerfile: ci/containers/Containerfile
+      target: prod
       build-args: USERID=1000
```
paulschreiber, line 27: Should we add tags here (or remove from FPF)?
```suggestion
      build-args: USERID=1000
      tags: ${{ needs.prepare.outputs.tags }}      
```

.github/workflows/check.yaml
```diff
@@ -51,24 +50,62 @@ jobs:
     uses: freedomofpress/actionslib/.github/workflows/oci-build.yaml@main
     with:
       context: '.'
-      containerfile: devops/docker/ProdDjangoDockerfile
-      build-args: USERID=12345
+      containerfile: ci/containers/Containerfile
+      target: prod
+      # No USERID override: the Containerfile defaults to 1000, matching both
+      # publish.yaml and the k8s securityContext, so the image tested here is
+      # the same artifact that gets deployed.
```
paulschreiber, line 57: Can we remove the override from FPF?
```
build-args: USERID=1000
```

prod-docker-compose.yaml
```diff
@@ -33,36 +34,36 @@ services:
...
   django:
+    init: true
     build:
       context: .
-      dockerfile: devops/docker/ProdDjangoDockerfile
-      args:
-        USERID: 12345
-    image: quay.io/freedomofpress/securedroporg
+      dockerfile: ci/containers/Containerfile
+      target: prod
```
paulschreiber, line 41: Can we remove `args: USERID: 1000` from FPF?

ci/containers/Containerfile
```diff
@@ -0,0 +1,196 @@
...
+
+# The previous Dev and Prod images installed the same set; Dev merely split it
+# across two stages. `gcc` is omitted as redundant -- build-essential depends on
+# it. The -dev headers are retained for any dependency that lacks a cp314 wheel
+# and falls back to an sdist.
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
+    curl \
+    git \
+    libffi-dev \
```
paulschreiber: SDO has `libffi-dev` but FPF does not. Do we need it?

justfile
```diff
@@ -0,0 +1,120 @@
...
+# Write the host UID into .env so containers build/run as your user (see README).
+dev-init:
+    echo "UID=$(id -u)" > .env
+
+# Same, but non-destructive: every compose invocation interpolates ${UID:?err},
+# so recipes depend on this to work from a cold checkout without clobbering an
+# .env the developer has added their own variables to.
+[private]
+env-check:
+    [ -f .env ] || echo "UID=$(id -u)" > .env
```
paulschreiber, line 34: should we add this `env-check` to FPF?

paulschreiber:
@conorsch looks like we'll need to change one of the required checks on the repo from `test:prod` to `test:production` in GitHub settings.

.github/workflows/check.yaml
conorsch: No, the opposite: we need to add it here. I'll make a comment in both places to make it clearer why we need it, but it boils down to supporting the multiplicity of container runtimes on developer laptops.

prod-docker-compose.yaml
conorsch: See above: no, we want it here, too—I'll add it.

ci/containers/Containerfile
conorsch: AFAICT `libffi-dev` was used for a pip build but our deps install fine without it. Will remove, and note that in the commit message.

.github/workflows/publish.yaml
conorsch: Good catch. I did some reading on the [actionslib](https://github.com/freedomofpress/actionslib/) logic, and I believe it's a no-op—evaluating to an empty string—so I've removed it in both places now.

conorsch:
> looks like we'll need to change one of the required checks on the repo from test:prod to test:production in GitHub settings.

You're right. That's a bit of a pain, but in the interest of achieving parity with the tooling across the repos, let's rip off the bandaid and get it done. IIRC it's actually a two step process:

1. _remove_ the old name from required checks 
2. _add_ the new name to required checks

I've done 1, but cannot do 2 until this PR lands, and the workflow with the revised name is on `develop`. We can follow-up post-merge.

justfile
```diff
@@ -0,0 +1,120 @@
...
+        coverage html --skip-empty --omit={{coverage_omit}}; \
+        coverage report -m --fail-under=70 --skip-empty --omit={{coverage_omit}}"
+
+# Inject development data into the postgresql database.
+createdevdata:
+    {{compose}} exec django bash -c "./manage.py createdevdata"
+
+# Import a postgres export file located at ./import.db.
+import-db:
+    {{compose}} exec -T postgresql bash -c "sed 's/OWNER TO [a-z]*/OWNER TO postgres/g' /django/import.db | psql securedropdb -U postgres > /dev/null"
```
conorsch, line 83: I don't use these db targets myself, so they're rather mechanical ports from the FPF repo. Open to feedback on changing. Also, we'll be forced to revisit in the near future, given the planned [pg14 -> pg17 migration](https://github.com/freedomofpress/infrastructure/issues/6886).

justfile
```diff
@@ -0,0 +1,120 @@
...
+# The final chown hands the regenerated lockfile back to whoever owns the input;
+# the builder runs as root, so without it a developer is left with root-owned
+# requirements files in their checkout.
+_pip-lock outfile infile *FLAGS:
+    {{engine}} run --rm -v "{{justfile_directory()}}:/code" -w /code {{python_builder}} \
+        bash -c 'apt-get update && apt-get install -y --no-install-recommends gcc libpq-dev && \
+            pip install pip-tools=={{pip_tools_version}} && \
+            pip-compile --generate-hashes --no-header --allow-unsafe {{FLAGS}} \
+                --output-file {{outfile}} {{infile}} && \
+            chown "$(stat -c "%u:%g" {{infile}})" {{outfile}}'
```
conorsch, line 110: Reminder that all this pip-tools madness will be revisited in a subsequent PR, once the dust settles on the container images.

justfile
conorsch: This comment I missed. Sure, could be added, but also fine to leave as-is. The important thing is that the CI checks are running against the dep files, and that folks can use the dev env locally.

conorsch:
Thanks for the deep review, @paulschreiber. I believe I've addressed everything called out, across this PR [and the corresponding one against fpf](https://github.com/freedomofpress/freedom.press/pull/3235). There's one unaddressed comment about an `env-check` just target, but it feels like we're hitting diminishing returns on quest for full parity, so I'm inclined to proceed.

paulschreiber (approved):

PR #1605, opened 2026-09-17
fix: add eslint and format JS files

### LINKED ISSUE #1291: Lint JavaScript with eslint
## Description
Lint JavaScript with eslint
- locally
- in CI

paulschreiber:
## Description

- Add an eslint configuration that matches PFT.
- Add eslint and prettier to pre-commit and CI.
- Add prettier to pre-commit. 
- Format JS files

Fixes https://github.com/freedomofpress/securedrop.org/issues/1291

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes


debug/static/debug/jquery.js
ryanwhitney: this file (vendored minified jQuery) should be ignored

package.json
```diff
@@ -1,47 +1,54 @@
...
+		"build": "webpack --mode production",
+		"start": "webpack watch --mode development --devtool eval-source-map",
+		"js-lint": "eslint client/**/*.js",
+		"stylelint": "stylelint \"client/**/*.scss\""
+	},
+	"author": "Harris Lapiroff <harris@littleweaverweb.com>",
+	"browserslist": [
+		"baseline widely available"
+	],
+	"devDependencies": {
```
ryanwhitney: Should prettier be in devDependencies so we're all on the same version?

package.json
```diff
@@ -1,47 +1,54 @@
...
-  }
+	"name": "securedrop",
+	"version": "0.0.0",
+	"private": true,
+	"description": "",
+	"main": "webpack.config.js",
+	"scripts": {
+		"build": "webpack --mode production",
+		"start": "webpack watch --mode development --devtool eval-source-map",
+		"js-lint": "eslint client/**/*.js",
```
ryanwhitney: same unquoted glob issue as we fixed for stylelint below. can either quote it or do `eslint .`

package.json
```diff
@@ -1,47 +1,54 @@
...
+		"@babel/preset-react": "^8.0.1",
+		"autoprefixer": "^10.5.4",
+		"babel-loader": "^10.1.1",
+		"babel-plugin-add-module-exports": "^1.0.4",
+		"css-loader": "^7.1.4",
+		"eslint": "^9.39.5",
+		"eslint-config-prettier": "^10.1.8",
+		"eslint-import-resolver-webpack": "^0.13.11",
+		"eslint-plugin-import": "^2.32.0",
+		"eslint-plugin-jsx-a11y": "^6.10.2",
```
ryanwhitney: are we importing `eslint-plugin-jsx-a11y` `eslint-plugin-react` and `eslint-plugin-react-hooks` because they may be called in a shared config? is there a way to not have these dependencies since they'll be unused?

package.json
ryanwhitney: we should also be pulling in `@eslint/js` and `globals` to ensure we're getting the version we want over transitive dependencies. (they are present in the PFT pr)

ryanwhitney (changes requested):
most of the actual js changes look good, just a few small notes on the overall setup

package.json
ryanwhitney: `eslint .` might be preferred here as it looks like the config files also declares scope

